PR #14, opened 2023-06-07
Eas 1592 - fixing the misclassification of functions in consumers

gersbach:
* ensures that functions exclusively in the classifier are not tied into vulns


.gitignore
```diff
@@ -1 +1,2 @@
 /target
+/test-apps
```
jwong101, line 2: This shouldn't be in the .gitignore, since the test-apps should be included in the repository.

.DS_Store
jwong101, line 1: Can you remove this file?

crates/forge_loader/src/manifest.rs
```diff
@@ -120,6 +120,21 @@ pub struct AppInfo<'a> {
     pub id: &'a str,
 }
 
+#[derive(Default, Debug, Clone, PartialEq, Eq, Deserialize)]
+pub struct ModuleResolve<'a> {
+    pub function: Option<&'a str>
```
jwong101: Does this need to be wrapped in an Option? AFAIK, the function attribute has to be present on all resolvers.

crates/forge_loader/src/manifest.rs
```diff
@@ -120,6 +120,21 @@ pub struct AppInfo<'a> {
...
+}
+
+#[derive(Default, Debug, Clone, PartialEq, Eq, Deserialize)]
+pub struct Module<'a> {
+    #[serde(default, borrow)]
+    pub function: Option<&'a str>,
+    #[serde(default, borrow)]
+    pub resolver: Option<ModuleResolve<'a>>,
```
jwong101, line 133: Maybe we should just reuse `Resolver`, but wrap the `method` field in an `Option`.

crates/forge_loader/src/manifest.rs
```diff
@@ -209,6 +224,25 @@ impl<'a> ForgeModules<'a> {
                     .map(|trigger| trigger.raw.function),
             )
             .collect();
+
+        let mut alternate_functions: Vec<&str> = Vec::new();
+        for module in self.extra {
```
jwong101, line 229:
```suggestion
        for module in self.extra.into_values().flatten() {
            alternate_functions.extend(module.function);
            if let Some(resolver) = module.resolver {
                // assuming this field is no longer wrapped in an Option
                alternate_functions.push(resolver.function);
            }
         }
```

.DS_Store
jwong101, line 1: Can you remove this file?

PR #15, opened 2023-06-13
Lowering Object LIterals

gersbach:
* Lowers object literals as operand


crates/forge_analyzer/src/definitions.rs
```diff
@@ -1102,8 +1102,82 @@ impl<'cx> FunctionAnalyzer<'cx> {
...
+                                                        def_id_prop,
+                                                    ));
+                                                    var.projections
+                                                        .push(Projection::Known(ident.sym.clone()));
+                                                }
+                                                _ => {}
+                                            }
+                                            self.body.push_inst(
+                                                self.block,
+                                                Inst::Assign(Variable::new(var_id), rval),
```
jwong101, line 1168:
```suggestion
                                                Inst::Assign(var, rval),
```

PR #16, opened 2023-06-23
Adding a constant propagation pass for definition analysis

gersbach:
* I was thinking this might be done, but after going through the code, I realize that there are a few things to do and clean up.


crates/forge_analyzer/src/reporter.rs
```diff
@@ -1,4 +1,6 @@
+use forge_loader::forgepermissions::ForgePermissions;
```
jwong101, line 1: Should probably remove these imports now.

crates/forge_analyzer/src/permissionclassifier.rs
```diff
@@ -0,0 +1,280 @@
...
+use forge_loader::forgepermissions::ForgePermissions;
+
+pub(crate) fn check_permission_used(
+    function_name: IntrinsicName,
+    first_arg: &str,
+    second_arg: Option<&str>,
+) -> Vec<ForgePermissions> {
+    let mut used_permissions: Vec<ForgePermissions> = Vec::new();
+
+    let post_call = second_arg.unwrap_or(&String::from("")).contains("POST");
```
jwong101, line 11:
```suggestion
    let post_call = second_arg.unwrap_or("").contains("POST");
```

You can also have `second_arg` just be a `&str` and pass in an empty string instead of `None`.

crates/forge_analyzer/src/checkers.rs
```diff
@@ -446,3 +459,597 @@ impl IntoVuln for AuthNVuln {
...
+fn find_member_of_obj(member: &str, obj: &Class) -> Option<DefId> {
+    for (mem, memdefid) in &obj.pub_members {
+        if mem == member {
+            return Some(memdefid.clone());
+        }
+    }
+    None
+}
+
+fn add_elements_to_intrinsic_struct(value: &Value, args: &mut Option<Vec<String>>) {
```
jwong101, line 929: It might be better to have `args` just be a `&mut Vec`, since this function calls `unwrap()` on it anyways.

crates/forge_analyzer/src/checkers.rs
```diff
@@ -446,3 +459,597 @@ impl IntoVuln for AuthNVuln {
...
+        Literal::Number(num) => Some(num.to_string()),
+        Literal::Str(str) => Some(str.to_string()),
+        _ => None,
+    }
+}
+
+fn find_member_of_obj(member: &str, obj: &Class) -> Option<DefId> {
+    for (mem, memdefid) in &obj.pub_members {
+        if mem == member {
+            return Some(memdefid.clone());
```
jwong101, line 923:
```suggestion
            return Some(memdefid);
```

crates/forge_analyzer/src/worklist.rs
```diff
@@ -75,15 +75,20 @@ where
 
 impl WorkList<DefId, BasicBlockId> {
     #[inline]
-    pub(crate) fn push_front_blocks(&mut self, env: &Environment, def: DefId) -> bool {
+    pub(crate) fn push_front_blocks(
```
jwong101, line 78: Is adding the arguments to each worklist task still necessary, since the current checker can just access the current arguments through the callstack?

crates/forge_analyzer/src/ir.rs
```diff
@@ -63,12 +64,12 @@ pub(crate) enum Terminator {
     },
 }
 
-#[derive(Clone, Debug, Copy, PartialEq, Eq)]
+#[derive(Clone, Debug, PartialEq, Eq)]
 pub enum Intrinsic {
-    Authorize,
+    Authorize(String),
```
jwong101, line 69: Should probably create an enum representing the type of API call instead of storing the raw `requestJira/requestConfluence`, since it's not clear what exactly the raw String is supposed to represent.

crates/forge_analyzer/src/worklist.rs
```diff
@@ -27,7 +27,7 @@ where
     }
 
     #[inline]
-    pub fn pop_front(&mut self) -> Option<(V, W)> {
+    pub fn pop_front(&mut self) -> (Option<(V, W)>) {
```
jwong101, line 30:
```suggestion
    pub fn pop_front(&mut self) -> Option<(V, W)> {
```

gersbach:
Closing because more updated code was pushed in

PR #18, opened 2023-07-10
Adding functionality for lowering functions that are called with handlers

gersbach:
This should take care of all functions that are called through handlers in JSX components, this is a little different than the approach we discussed in slack, but it looks good from my end, and I should be able to extract permissions more accurately :). Lmk what you think.

### DIFF
--- a/crates/forge_analyzer/src/definitions.rs
+++ b/crates/forge_analyzer/src/definitions.rs
@@ -1120,9 +1120,7 @@ impl<'cx> FunctionAnalyzer<'cx> {
                 let var = self.body.get_or_insert_global(def);
                 Operand::with_var(var)
             }
-            JSXElementName::JSXMemberExpr(mem) => {
-                self.lower_jsx_member(&mem)
-            }
+            JSXElementName::JSXMemberExpr(mem) => self.lower_jsx_member(&mem),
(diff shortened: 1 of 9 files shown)


crates/forge_analyzer/src/definitions.rs
```diff
@@ -1021,6 +1024,49 @@ impl<'cx> FunctionAnalyzer<'cx> {
...
+                }
+            }
+            JSXAttrOrSpread::SpreadElement(_) => {}
+        });
+    }
+
+    fn lower_jsx_handler(&mut self, n: &JSXExprContainer, ident_value: &Ident) {
+        if let JSXExpr::Expr(expr) = &n.expr {
+            self.lower_expr(&expr);
+            if ident_value.sym.contains("on") {
```
jwong101, line 1045: To make this more accurate, you could make sure the ident starts with "on" and the next letter is capitalized.

crates/forge_analyzer/src/definitions.rs
```diff
@@ -1021,6 +1024,49 @@ impl<'cx> FunctionAnalyzer<'cx> {
...
+                match &**expr {
```
jwong101, line 1046: Make sure to add a "TODO" or "FIXME" to eventually add this as an entrypoint.

jwong101 (commented):
Just this one change and it lgtm.

PR #19, opened 2023-07-20
Handling  default and old nodeJS exports

gersbach:
* Finally found the bug :)


crates/forge_analyzer/src/definitions.rs
```diff
@@ -2165,6 +2209,43 @@ impl Visit for ExportCollector<'_> {
         };
     }
 
+    fn visit_assign_expr(&mut self, n: &AssignExpr) {
+        if let Some(ident) = ident_from_assign_expr(n) {
+            if ident.sym.to_string() == "module" {
+                if let Some(mem_expr) = mem_expr_from_assign(n.clone()) {
+                    if let MemberProp::Ident(ident_property) = &mem_expr.prop {
+                        if ident_property.sym.to_string() == "exports" {
```
jwong101, line 2217:
```suggestion
                        if ident_property.sym == "exports" {
```

crates/forge_analyzer/src/definitions.rs
```diff
@@ -2165,6 +2209,43 @@ impl Visit for ExportCollector<'_> {
...
+                                Expr::Class(ClassExpr { ident, class }) => self.add_default(
+                                    DefRes::Class(()),
+                                    ident.as_ref().map(Ident::to_id),
+                                ),
+                                _ => {}
+                            }
+                        }
+                    }
+                }
+            } else if ident.sym.to_string() == "exports" {
```
jwong101, line 2232:
```suggestion
            } else if ident.sym == "exports" {
```

crates/forge_analyzer/src/definitions.rs
```diff
@@ -1540,10 +1540,15 @@ impl Visit for LocalDefiner<'_> {
...
-impl Visit for FunctionCollector<'_> {
-    fn visit_function(&mut self, n: &Function) {
-        n.visit_children_with(self);
+impl FunctionCollector<'_> {
+    fn handle_function(&mut self, n: &Function, owner: Option<DefId>) {
         let owner = self.parent.unwrap_or_else(|| {
+            if let Some(defid) = owner {
+                return defid;
+            }
+            if let Some(defid) = self.res.default_export(self.module) {
```
jwong101, line 1549: Won't this alias closures in modules with a default export? You shouldn't need to check this case here, since the owner should be set to the default export if this function is called on an assign expression to `exports` or `module.exports`.

crates/forge_analyzer/src/definitions.rs
```diff
@@ -1579,6 +1584,40 @@ impl Visit for FunctionCollector<'_> {
             *self.res.def_mut(owner).expect_body() = body;
         }
     }
+}
+
+impl Visit for FunctionCollector<'_> {
```
jwong101, line 1589: Also, I just realized you'll have to set the owner correctly for stuff like:
`export default function() {}` where the function doesn't have a name.

PR #22, opened 2023-07-24
Lowering then and array functions

gersbach:
I'll be moving a lot of other functions to the utils file from the other files to keep the code clean if this sounds to good to you

### DIFF
--- a/crates/forge_analyzer/src/definitions.rs
+++ b/crates/forge_analyzer/src/definitions.rs
@@ -2,7 +2,7 @@
 
 use std::{borrow::Borrow, fmt, mem};
 
-use crate::utils::call_func_with_name;
+use crate::utils::calls_method;
 use forge_file_resolver::{FileResolver, ForgeResolver};
 use forge_utils::{create_newtype, FxHashMap};
 
@@ -936,9 +936,13 @@ impl<'cx> FunctionAnalyzer<'cx> {
             if self.res.as_foreign_import(id, "@forge/ui").map_or(
                 false,
                 |imp| matches!(imp, ImportKind::Named(s) if *s == *"useState"),
-            ) {
+            ) || calls_method(callee, "then")
+                || calls_method(callee, "map")
+                || calls_method(callee, "foreach")
+                || calls_method(callee, "filter")
+            {
                 if let [ExprOrSpread { expr, .. }] = args {
-                    debug!("found useState");
+                    debug!("found useState/then/map/foreach/filter");
                     match &**expr {
                         Expr::Arrow(ArrowExpr { body, .. }) => match body {
                             BlockStmtOrExpr::BlockStmt(stmt) => {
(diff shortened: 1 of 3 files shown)


jwong101:
Can you do this in the `lower_calls` function instead here: https://github.com/atlassian-labs/FSRT/blob/2b8c2fcbc057869c597ab7a1c63f09a953127a94/crates/forge_analyzer/src/definitions.rs#L939? It should be similar to what happens in useState.

crates/forge_analyzer/src/definitions.rs
```diff
@@ -959,6 +960,25 @@ impl<'cx> FunctionAnalyzer<'cx> {
                 }
             }
         }
+
+        if call_func_with_name(&callee, "then")
+            || call_func_with_name(&callee, "map")
+            || call_func_with_name(&callee, "foreach")
+            || call_func_with_name(&callee, "filter")
+        {
+            if let Some(lambda_function) = args.get(0) {
```
jwong101, line 969: Can you add this check to the if statement above? That way, this will handle other stuff like non block statement expressions in arrow functions and function expressions.

crates/forge_analyzer/src/utils.rs
```diff
@@ -0,0 +1,15 @@
+use crate::definitions::CalleeRef;
+use swc_core::ecma::ast::{CallExpr, Callee, Expr, MemberProp};
+
+pub fn call_func_with_name(n: &CalleeRef, name: &str) -> bool {
```
jwong101, line 4:
```suggestion
pub fn calls_method(n: CalleeRef<'_>, name: &str) -> bool {
```
`CalleeRef` is essentially already a reference, so it's already `Copy`. So passing in a reference to `CalleeRef` is just double indirection.

crates/forge_analyzer/src/utils.rs
```diff
@@ -0,0 +1,15 @@
...
+    if let CalleeRef::Expr(expr) = &n {
```
jwong101, line 5:
```suggestion
    if let CalleeRef::Expr(Expr::Member(mem)) = n {
```
You can take advantage of [match ergonomics](https://rust-lang.github.io/rfcs/2005-match-ergonomics.html).

crates/forge_analyzer/src/utils.rs
```diff
@@ -0,0 +1,15 @@
...
+        if let Expr::Member(mem) = &**expr {
+            if let MemberProp::Ident(ident) = &mem.prop {
+                if ident.sym.to_string() == name {
```
jwong101, line 8:
```suggestion
    return ident.sym == *name;
```
Since `JsWord` impls `PartialEq<str>`. Also, the reason you have to dereference `name` is because `lhs == rhs` in Rust essentially just desugars to `PartialEq::eq(&lhs, &rhs)`. Thus, `ident.sym == name` will try to find an `impl PartialEq<&str>`(which doesn't exist), since it's being passed `PartialEq::eq(&ident.sym, &name)`.

jwong101:
LGTM

PR #23, opened 2023-07-24
Prevents scan if the code is transpiled

gersbach:
* This is required for my scanner because we add all permissions initially, and when we find urls we remove the permissions as we go. However, if we can't scan any of the code, then we cannot extract any permissions and creates false positives. This is more efficient because there is no point in scanning anyway if the code is transpiled and cannot be scanned to begin with.

### DIFF
--- a/crates/fsrt/src/main.rs
+++ b/crates/fsrt/src/main.rs
@@ -177,17 +177,15 @@ fn scan_directory(dir: PathBuf, function: Option<&str>, opts: Opts) -> Result<Fo
 
     let paths = collect_sourcefiles(dir.join("src/")).collect::<HashSet<_>>();
 
-    let transpiled_async = paths
-        .iter()
-        .map(|path| {
-            if let Ok(data) = fs::read_to_string(path) {
-                if let Some(line) = data.lines().nth(0) {
-                    return line.contains("\"use strict\";");
-                }
-            }
-            false
-        })
-        .any(|path| path);
+    let transpiled_async = paths.iter().any(|path| {
+        if let Ok(data) = fs::read_to_string(path) {
+            return data
+                .lines()
+                .next()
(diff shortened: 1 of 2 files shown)


crates/fsrt/src/main.rs
```diff
@@ -175,6 +176,19 @@ fn scan_directory(dir: PathBuf, function: Option<&str>, opts: Opts) -> Result<Fo
     let name = manifest.app.name.unwrap_or_default();
 
     let paths = collect_sourcefiles(dir.join("src/")).collect::<HashSet<_>>();
+
+    let transpiled_async = paths
+        .iter()
+        .map(|path| {
```
jwong101, line 182:
```suggestion
        .any(|path| {
```

crates/fsrt/src/main.rs
```diff
@@ -175,6 +176,19 @@ fn scan_directory(dir: PathBuf, function: Option<&str>, opts: Opts) -> Result<Fo
...
+            if let Ok(data) = fs::read_to_string(path) {
+                if let Some(line) = data.lines().nth(0) {
```
jwong101, line 184:
```suggestion
                if let Some(line) = data.lines().next() {
```

crates/fsrt/src/main.rs
```diff
@@ -175,6 +176,19 @@ fn scan_directory(dir: PathBuf, function: Option<&str>, opts: Opts) -> Result<Fo
...
+                    return line.contains("\"use strict\";");
```
jwong101, line 185: You might want to check for `'use strict'` as well, i.e. if they use single quotes. Also, you can use `https://doc.rust-lang.org/stable/std/option/enum.Option.html#method.is_some_and`, but you'll have to bump the docker image to use rustc 1.70 or higher.

PR #26, opened 2023-08-08
Secret Scanning for Forge Apps

gersbach:


gersbach:
Don't merge this in yet, I still need to work on finishing all of the cases for the definition analysis

PR #28, opened 2023-11-06
Eas 1965

aliner-wang:
Added new pathway for scanning for packages requiring secrets to include * imported packages. 

Updated secretdata.yaml with new packages to scan for. Commented out packages use require import method that we don't scan for. 

Added package import to DVFA.

Todo: definitely can clean up code from comments and stuff in DVFA and around main. But as_intrinsic in definition.rs is done and cleaned up.


secretdata.yaml
```diff
@@ -0,0 +1,50 @@
+# Data Structure explained
+# - package_name: name of package being checked
+#   identifier: type of import
+#   function_name: name of the function that takes the secret data
+#   secret_position: which argument takes in the secret data
+
+- package_name: jsonwebtoken
+  identifier: default
+  function_name: Some
```
jwong101, line 9: shouldn't this be `sign`?

crates/forge_analyzer/src/definitions.rs
```diff
@@ -469,12 +548,35 @@ struct ExportCollector<'cx> {
...
+pub enum IntrinsicName {
+    RequestConfluence,
+    RequestJira,
+    Other,
+}
+
+#[derive(Serialize, Deserialize, PartialEq, Debug, Clone, Eq)]
+pub struct PackageData {
+    pub package_name: String,
+    pub identifier: String,
```
jwong101, line 575: I think this should be the actual identifier that's exported. For `default` exports, this makes sense, because `export default` technically names the exported object or function "default". However, "star" should technically just be the actual function or object identifier, since the package itself doesn't really care if its being imported as a named or star import. For instance:

```javascript
import * as jwt from 'jsonwebtoken'
jwt.sign('foo','secret');

//can also be imported like this:
import { sign } from 'jsonwebtoken';
sign('foo', 'secret');
```
So the identifier should simply be 'sign'. By keeping track of the identifier there, we can also support objects like in `crypto-js` by renaming `function_name` to `method` and having it be an `Option<String>`. So if we wanted to support `AES.encrypt` (which is an object) in `crypto-js`, we would have:

```yaml
package_name: crypto-js
identifier: AES
method: encrypt
secret_position: 2
```

crates/forge_analyzer/src/definitions.rs
```diff
@@ -787,29 +920,34 @@ impl<'cx> FunctionAnalyzer<'cx> {
         fn is_storage_read(prop: &JsWord) -> bool {
             *prop == *"get" || *prop == *"getSecret" || *prop == *"query"
         }
-
         match *callee {
             [PropPath::Unknown((ref name, ..))] if *name == *"fetch" => Some(Intrinsic::Fetch),
+            // [PropPath::Def(def)] => {}]
```
jwong101, line 925: Don't forget to handle the case where it's just a "regular" function call, i.e. with just an identifier. Otherwise, we wouldn't catch this:

```javascript
import { sign } from 'jsonwebtoken';
sign('foo', 'secret');
```

test-apps/jira-damn-vulnerable-forge-app/src/utils.js
```diff
@@ -1,21 +1,119 @@
...
+  func_from_exports();
+  another_export();
+  newExport();
+  func_defult();
+  my_function();
+  let global = 'test';
+
+  // testing all the libraries
+  // var token = jwt.sign({ foo: 'bar' }, 'peek a boo');
+  // var hmac = HmacMD5('Secret Message', 'HMAC PASSWORD');
```
jwong101, line 76: Can you add test cases for named imports as well? Also for cases where an identifier could shadow an import to see if we don't raise a false positive there.

crates/forge_analyzer/src/definitions.rs
jwong101: Yeah, that would return `false` if it's not an import or if it's imported from a different project.

crates/forge_analyzer/src/definitions.rs
```diff
@@ -2981,11 +3503,42 @@ impl Environment {
...
+            ImportKind::Star => true,
+            ImportKind::Default => true,
+            ImportKind::Named(ref name) => {
+                if let Some(import_name) = named_import {
+                    return *name == *import_name;
+                } else {
+                    false
+                }
+            }
+            _ => false,
```
jwong101, line 3518: You shouldn't need this catchall case.

crates/forge_analyzer/src/definitions.rs
```diff
@@ -2981,11 +3503,42 @@ impl Environment {
             DefKind::Undefined => DefKind::Undefined,
         }
     }
+    pub fn is_import(&self, import_kind: ImportKind, named_import: Option<String>) -> bool {
```
jwong101, line 3506: This should probably just be a method an `ImportKind`, since we don't use the resolver here. Also, this could probably just take in a `&str` and check if the import is the same identifier as what was passed in.

crates/forge_analyzer/src/definitions.rs
```diff
@@ -2981,11 +3503,42 @@ impl Environment {
...
+        // debug!("{named_import}");
+        match import_kind {
+            ImportKind::Star => true,
+            ImportKind::Default => true,
```
jwong101, line 3510: This should be true iff `named_import` is "default", since this function is for checking if the import matches whatever identifier/string we pass in.

crates/forge_analyzer/src/definitions.rs
```diff
@@ -31,12 +33,37 @@ use swc_core::{
...
         atoms::{Atom, JsWord},
+        utils::function,
         visit::{noop_visit_type, Visit, VisitWith},
     },
 };
-use tracing::{debug, info, instrument, warn};
+use tracing::{debug, field::debug, info, instrument, warn};
 use typed_index_collections::{TiSlice, TiVec};
 
+/**
```
jwong101, line 43: Can you remove these macros?

crates/forge_analyzer/src/definitions.rs
```diff
@@ -819,21 +956,119 @@ impl<'cx> FunctionAnalyzer<'cx> {
...
+                } else {
+                    None
+                }
+            }
+            // [Def, .., Static] star
+            // [Def, .., Static] named/default
+            // [PropPath::Def(def)]
+            // 1. Named/Default => package matches imported from and ident is the same, (NO METHOD)
+            [PropPath::Def(def)] if self.res.as_foreign_import(def).is_some() => {
+                if let Some((package_name, import_kind)) = self.res.as_foreign_import(def) {
```
jwong101, line 1047:
```suggestion
                let (package_name, import_kind) = self.res.as_foreign_import(def)?;
```

crates/forge_analyzer/src/definitions.rs
```diff
@@ -819,21 +956,119 @@ impl<'cx> FunctionAnalyzer<'cx> {
...
+                    }
+                } else {
+                    None
+                }
+            }
+            // [Def, .., Static] star
+            // [Def, .., Static] named/default
+            // [PropPath::Def(def)]
+            // 1. Named/Default => package matches imported from and ident is the same, (NO METHOD)
+            [PropPath::Def(def)] if self.res.as_foreign_import(def).is_some() => {
```
jwong101, line 1046: We should be able to remove this guard as long as this match branch happens near the end, since you have to lookup the `package_name` and `import_kind` in the match arm's block using this function anyway.

crates/forge_analyzer/src/definitions.rs
```diff
@@ -819,21 +956,119 @@ impl<'cx> FunctionAnalyzer<'cx> {
...
+                if let Some((package_name, import_kind)) = self.res.as_foreign_import(def) {
+                    let package = self.secret_packages.iter().find(|&package_data| {
+                        *package_name == *package_data.package_name
+                            && import_kind == *package_data.identifier
+                            && package_data.method.is_none()
+                    });
+                    if package.is_some() {
```
jwong101, line 1053:
```suggestion
                    package.cloned().map(Intrinsic::JWTSign)
```
See [Option::cloned](https://doc.rust-lang.org/std/option/enum.Option.html#method.cloned) and [Option::map](https://doc.rust-lang.org/std/option/enum.Option.html#method.map). Note that tuple structs and enum tuple variants are essentially functions.

crates/forge_analyzer/src/ir.rs
```diff
@@ -67,14 +70,15 @@ pub(crate) enum Terminator {
...
-    Authorize,
+    Authorize(IntrinsicName),
     Fetch,
-    ApiCall,
-    SafeCall,
     UserFieldAccess,
     ApiCustomField,
+    ApiCall(IntrinsicName),
+    SafeCall(IntrinsicName),
+    JWTSign(PackageData),
```
jwong101, line 81: Can you rename this variant to `SecretFunction` or something more general? Also, don't forget to use VSCode's rename symbol function that's on `F2`.

crates/forge_analyzer/src/interp.rs
```diff
@@ -556,16 +854,16 @@ impl<'cx, C: Checker<'cx>> Interp<'cx, C> {
...
-            writeln!(out, "No vulnerabilities found")
-        } else {
-            writeln!(out, "Found {} vulnerabilities", vulns.len())?;
-            for vuln in vulns {
-                writeln!(out, "{vuln}")?;
-            }
-            Ok(())
-        }
-    }
+    // pub fn dump_results(&self, out: &mut dyn Write) -> io::Result<()> {
```
jwong101, line 857: Can you delete this?

crates/forge_analyzer/src/checkers.rs
```diff
@@ -105,28 +120,32 @@ impl<'cx> Dataflow<'cx> for AuthorizeDataflow {
...
 pub struct AuthZChecker {
+    visit: bool,
     vulns: Vec<AuthZVuln>,
 }
 
 impl AuthZChecker {
     pub fn new() -> Self {
-        Self { vulns: vec![] }
+        Self {
+            visit: true,
```
jwong101, line 146: Can you delete this field? It's not used anywhere.

crates/forge_analyzer/src/checkers.rs
```diff
@@ -471,3 +510,1121 @@ impl IntoVuln for AuthNVuln {
...
+
+impl PermissionChecker {
+    pub fn new() -> Self {
+        Self {
+            visit: false,
+            vulns: vec![],
+        }
+    }
+
+    pub fn into_vulns(self, permissions: Vec<String>) -> impl IntoIterator<Item = PermissionVuln> {
```
jwong101, line 1022:
```suggestion
    pub fn into_vulns(self, unused_permissions: Vec<String>) -> impl IntoIterator<Item = PermissionVuln> {
```

crates/forge_analyzer/src/checkers.rs
```diff
@@ -471,3 +510,1121 @@ impl IntoVuln for AuthNVuln {
...
+        if permissions.len() > 0 {
+            return Vec::from([PermissionVuln {
+                unused_permissions: permissions.clone(),
```
jwong101, line 1025:
```suggestion
                unused_permissions,
```

crates/forge_analyzer/src/definitions.rs
```diff
@@ -438,21 +514,24 @@ impl Copy for DefRes {}
...
     exports: TiVec<ModId, Vec<(JsWord, DefId)>>,
-    defs: Definitions,
+    pub global: TiVec<ModId, DefId>,
+    pub defs: Definitions,
     default_exports: FxHashMap<ModId, DefId>,
-    resolver: ResolverTable,
+    pub resolver: ResolverTable,
 }
 
+// POI
```
jwong101, line 533: Can you delete this comment?

crates/forge_analyzer/src/definitions.rs
```diff
@@ -1077,7 +1322,7 @@ impl<'cx> FunctionAnalyzer<'cx> {
         match n {
             Pat::Ident(BindingIdent { id, .. }) => {
                 let id = id.to_id();
-                let def = self.res.get_or_insert_sym(id, self.module);
+                let def = self.res.get_or_insert_sym(id.clone(), self.module);
```
jwong101, line 1325:
```suggestion
                let def = self.res.get_or_insert_sym(id, self.module);
```

crates/forge_analyzer/src/definitions.rs
```diff
@@ -2617,20 +3037,36 @@ impl Visit for ExportCollector<'_> {
...
+                                        self.default.unwrap(),
+                                    );
+                                }
                                 _ => {}
                             }
                         }
                     }
                 }
-            } else if &ident.sym == "exports" {
+            } else if ident.sym.to_string() == "exports" {
```
jwong101, line 3053:
```suggestion
            } else if ident.sym == "exports" {
```

crates/forge_analyzer/src/definitions.rs
```diff
@@ -2678,19 +3116,34 @@ impl Visit for ExportCollector<'_> {
...
 
-    fn visit_export_default_expr(&mut self, _: &ExportDefaultExpr) {
-        self.add_default(DefRes::Undefined, None);
+    fn visit_export_default_expr(&mut self, n: &ExportDefaultExpr) {
+        if let Expr::Ident(ident) = &*n.expr {
+            self.add_default(DefRes::Undefined, None);
+            // adding the default export, so we can resolve it during the lowering
+            self.res_table
+                .exported_names
+                .insert((ident.sym.clone(), self.curr_mod), self.default.unwrap());
```
jwong101, line 3145: Could you change the `ExportCollector::set_default` method to return the `DefId` that was set to `self.default` at the end of that function?; That way you can do `let default_id = self.add_default(DefRes::Undefined, None);` and replace `self.default.unwrap()` with `default_id`.

crates/forge_analyzer/src/definitions.rs
```diff
@@ -2617,20 +3037,36 @@ impl Visit for ExportCollector<'_> {
                                     DefRes::Class(()),
                                     ident.as_ref().map(Ident::to_id),
                                 ),
+                                Expr::Ident(ident) => {
+                                    self.add_default(DefRes::Undefined, None);
+                                    // adding the default export, so we can resolve it during the lowering
+                                    self.res_table.exported_names.insert(
+                                        (ident.sym.clone(), self.curr_mod),
+                                        self.default.unwrap(),
```
jwong101, line 3045: Could you change the ExportCollector::set_default method to return the DefId that was set to self.default at the end of that function?; That way you can do let default_id = self.add_default(DefRes::Undefined, None); and replace self.default.unwrap() with default_id.

crates/forge_analyzer/src/definitions.rs
```diff
@@ -2592,22 +3011,23 @@ impl Visit for ExportCollector<'_> {
...
     }
 
     fn visit_assign_expr(&mut self, n: &AssignExpr) {
         if let Some(ident) = ident_from_assign_expr(n) {
-            if &ident.sym == "module" {
+            if ident.sym.to_string() == "module" {
                 if let Some(mem_expr) = mem_expr_from_assign(n) {
                     if let MemberProp::Ident(ident_property) = &mem_expr.prop {
-                        if &ident_property.sym == "exports" {
+                        if ident_property.sym.to_string() == "exports" {
```
jwong101, line 3030:
```suggestion
                        if ident_property.sym == "exports" {
```

crates/forge_analyzer/src/definitions.rs
```diff
@@ -819,21 +956,119 @@ impl<'cx> FunctionAnalyzer<'cx> {
...
+            }
+            // Case 1: the “root” object is a star import and the Static proppath member is either identifier or method
+            // [Def, Static, Static]
+            //[PropPath::Def(def), PropPath::Static(a), PropPath::Static(b)]
+            // 1. Star, identifier, method
+            [PropPath::Def(def), PropPath::Static(ref identifier), PropPath::Static(ref method)] => {
+                // case where function requires object to be invoked first
+                // example: import * as cryptoJS from 'crypto-js';
+                // var aes = cryptoJS.AES.encrypt('Secret message', 'secret password');
+                if let Some((package_name, import_kind)) = self.res.as_foreign_import(def) {
```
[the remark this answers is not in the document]
jwong101: See [this](https://doc.rust-lang.org/book/ch09-02-recoverable-errors-with-result.html#a-shortcut-for-propagating-errors-the--operator) for an explanation of how this works.

crates/forge_analyzer/src/definitions.rs
```diff
@@ -819,21 +956,119 @@ impl<'cx> FunctionAnalyzer<'cx> {
...
+                    });
+                    if package.is_some() {
+                        return Some(Intrinsic::JWTSign(package.unwrap().clone()));
+                    }
+                }
+
+                None
+            }
+
             [PropPath::Def(def), PropPath::Static(ref s), ..] if is_storage_read(s) => {
```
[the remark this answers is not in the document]
aliner-wang: ran into a problem with this one. Moving the last case above all the secret scanner checks led to the secret scanner checks to never get caught.

crates/forge_analyzer/src/definitions.rs
```diff
@@ -960,6 +960,21 @@ impl<'cx> FunctionAnalyzer<'cx> {
...
+                    }
+                    _ => None,
+                }
+            }
+            [PropPath::Def(def), ..] => match self.res.is_imported_from(def, "@forge/api") {
+                Some(ImportKind::Named(ref name)) if *name == *"authorize" => {
+                    Some(Intrinsic::Authorize(IntrinsicName::Other))
+                }
+                _ => None,
+            },
```
aliner-wang, line 977: Specifically moving this case above the secret scanner checks I think catches all statements before secret scanner checks.

jwong101: Ah, mb I thought that match was a guard statement for some reason. If you change it to a guard statement with `if self.res.is_imported_from(def, "@forge/api") == *"authorize" => Some(Intrinsic::Authorize(IntrinsicName::Other)),`, then moving it above shouldn't swallow the secret scanner checks.

crates/forge_analyzer/src/definitions.rs
aliner-wang: There is no set_default function.

aliner-wang: Tldr on my efforts: For this one, I couldn't find `set_default` under ExportCollector. Nor did a `default_id` variable exist. 

So I tried to return defid in add_default. But that also messes up some self.add_default(DefRes::Undefined, None); implementations elsewhere because it's expected that nothing gets returned. 

I did try creating self.set_default, but it seemed a little repetitive to call it if add_default is already setting self.default. and calling set_default to return a defId add_default is pretty much the same result as the approach described above. 

Otherwise, the easiest way to not use .unwrap() might be to use a let statement grab the value in a Some() object instead of returning it. 

Do you want me to break a lil and adapt like other places where self.add_default is called?

jwong101: Ah, I meant change `add_default` to return a DefId. Yeah, that should be fine, since looking at the breakage, it's just match arms that are now returning a `DefId` instead of `()`. To fix it, you just need to wrap the arm in braces and add a semicolon.

PR #31, opened 2024-01-05
Eas-1673

aliner-wang:
Added logic to track and test that jira:adminPage gets tracked as a admin module.


crates/forge_loader/src/manifest.rs
```diff
@@ -393,7 +393,7 @@ impl<'a> ForgeModules<'a> {
                 .webtriggers
                 .binary_search_by_key(&func.key, |trigger| &trigger.function)
                 .is_ok();
-            let invokable = non_user_invokable_mod_functions.contains(func.key);
+            let invokable = invokable_functions.contains(func.key);
```
aliner-wang, line 396: This part has me a little confused in why we do what we do. With the implementation as is, the scanner will run an AuthZ scan on any function manually tracked from a non-user invokable module. But wouldn't an AuthZ scan need to be run over all user-invocable functions? Because the case missed is for functions exposed in both types of modules. So I'm wondering where an AuthZ scan is performed on user-invokable module functions? And if they're necessary? 

Since the rest of the modules are presumed user-invokable, the compass modules with resolver traits should be scanned ?

jwong101: The implementation(on main) currently will consider a function as `invokable` iff it is not on any `non-invokable` functions. With these changes, we want to consider a function as `invokable` iff it exists on an `invokable` module(since that more closely models reality.

If a function is exposed by both types of modules, then the latter logic has the correct semantics, as we need to scan that function, since it's user invokable. The current logic(on main) doesn't handle this case correctly and ignores that that function, even though it is invokable.

Yeah, the compass modules we talked about should be invokable(by adding it the list of invokable functions). That's the one downside of this approach; we need to track every user invokable module, whereas before we only needed to track non user invokable modules(of which there are far fewer). However, it's necessary to handle the above edge case.

crates/forge_loader/src/manifest.rs
```diff
@@ -59,25 +89,66 @@ enum Interval {
...
 
+// Add more structs here for deserializing forge modules
 #[derive(Default, Debug, Clone, PartialEq, Eq, Deserialize)]
 pub struct ForgeModules<'a> {
+    // deserializing non user-invocable modules
     #[serde(rename = "macro", default, borrow)]
     macros: Vec<MacroMod<'a>>,
     #[serde(rename = "function", default, borrow)]
     pub functions: Vec<FunctionMod<'a>>,
+    #[serde(rename = "contentByLineItem", default, borrow)]
```
jwong101, line 145:
```suggestion
    #[serde(rename = "confluence:contentByLineItem", default, borrow)]
```

Also, you're missing these confluence modules:

- [`confluence:contentAction`](https://developer.atlassian.com/platform/forge/manifest-reference/modules/confluence-content-action/#properties)
- [`confluence:contextMenu`](https://developer.atlassian.com/platform/forge/manifest-reference/modules/confluence-context-menu/#properties)
- [`confluence:globalPage`](https://developer.atlassian.com/platform/forge/manifest-reference/modules/confluence-global-page/#properties)
- [`confluence:homepageFeed`](https://developer.atlassian.com/platform/forge/manifest-reference/modules/confluence-homepage-feed/#properties)
- [`confluence:spacePage`](https://developer.atlassian.com/platform/forge/manifest-reference/modules/confluence-space-page/#properties)
- [`confluence:spaceSettings`](https://developer.atlassian.com/platform/forge/manifest-reference/modules/confluence-space-settings/)

crates/forge_loader/src/manifest.rs
```diff
@@ -86,11 +157,28 @@ pub struct ForgeModules<'a> {
...
+    pub data_provider: Vec<DataProvider<'a>>,
+    #[serde(rename = "jira:customField", default, borrow)]
+    pub custom_field: Vec<CustomField<'a>>,
+    #[serde(rename = "jira:uiModificatons", default, borrow)]
+    pub ui_modifications: Vec<UiModificatons<'a>>,
+    #[serde(rename = "jira:workflowValidator", default, borrow)]
+    pub workflow_validator: Vec<WorkflowValidator<'a>>,
     #[serde(rename = "jira:workflowPostFunction", default, borrow)]
-    workflow_post_functions: Vec<WorkflowPostFunction<'a>>,
+    pub workflow_post_function: Vec<WorkflowPostFunction<'a>>,
```
jwong101, line 169: You're missing these jira modules:

- [`jira:globalPage`](https://developer.atlassian.com/platform/forge/manifest-reference/modules/jira-global-page/#jira-global-page)
- [`jira:issuePanel`](https://developer.atlassian.com/platform/forge/manifest-reference/modules/jira-issue-panel/#properties)
- [`jira:issueAction`](https://developer.atlassian.com/platform/forge/manifest-reference/modules/jira-issue-action/#properties)
- [`jira:issueContext`](https://developer.atlassian.com/platform/forge/manifest-reference/modules/jira-issue-context/#properties)
- [`jira:issuePanel`](https://developer.atlassian.com/platform/forge/manifest-reference/modules/jira-issue-panel/#properties)
- [`jira:issueViewBackgroundScript`](https://developer.atlassian.com/platform/forge/manifest-reference/modules/jira-issue-view-background-script/#properties)
- [`jira:projectSettingsPage`](https://developer.atlassian.com/platform/forge/manifest-reference/modules/jira-project-settings-page/#properties)
- [`jira:projectPage`](https://developer.atlassian.com/platform/forge/manifest-reference/modules/jira-project-page/#properties)
- [`jira:projectSettingsPage`](https://developer.atlassian.com/platform/forge/manifest-reference/modules/jira-project-settings-page/#properties)
- [`jira:issueActivity`](https://developer.atlassian.com/platform/forge/manifest-reference/modules/jira-issue-activity/#properties)
- [`jira:jqlFunction`](https://developer.atlassian.com/platform/forge/manifest-reference/modules/jql-function/#properties)
- [`jira:dashboardBackgroundScript`](https://developer.atlassian.com/platform/forge/manifest-reference/modules/jira-dashboard-background-script/#properties)
- [`jira:dashboardGadget`](https://developer.atlassian.com/platform/forge/manifest-reference/modules/jira-dashboard-gadget/#properties)
  Note that there's an optional invokable function at the `edit` property for this one.
- [`customFieldType`](https://developer.atlassian.com/platform/forge/manifest-reference/modules/jira-custom-field-type/#properties)
  Note that there's a *lot* of invokable functions on this module, so read through them carefully. The only non invokable one is the `searchSuggestions` function.

crates/forge_loader/src/manifest.rs
```diff
@@ -86,11 +157,28 @@ pub struct ForgeModules<'a> {
     scheduled_triggers: Vec<ScheduledTrigger<'a>>,
     #[serde(rename = "consumer", default, borrow)]
     pub consumers: Vec<Consumer<'a>>,
+    #[serde(rename = "compass:dataProvider", default, borrow)]
```
jwong101, line 160: The following compass modules are missing:

- [`compass:componentPage`](https://developer.atlassian.com/platform/forge/manifest-reference/modules/compass-component-page/#properties)
- [`compass:globalPage`](https://developer.atlassian.com/platform/forge/manifest-reference/modules/compass-global-page/#properties)
- [`compass:teamPage`](https://developer.atlassian.com/platform/forge/manifest-reference/modules/compass-team-page/#properties)
- [`compass:adminPage`](https://developer.atlassian.com/platform/forge/manifest-reference/modules/compass-admin-page/#properties)

crates/forge_loader/src/manifest.rs
```diff
@@ -208,63 +282,102 @@ impl<T> AsRef<T> for FunctionTy<T> {
...
+            invokable_functions.extend(customfield.common_keys.resolver);
+        });
+
+        self.ui_modifications.iter().for_each(|ui| {
+            invokable_functions.insert(ui.common_keys.function);
+            invokable_functions.extend(ui.common_keys.resolver);
+        });
+
+        self.workflow_validator.iter().for_each(|validator| {
+            invokable_functions.insert(validator.common_keys.key);
```
jwong101, line 319: The `key` property doesn't point to a function.

crates/forge_loader/src/manifest.rs
```diff
@@ -208,63 +282,102 @@ impl<T> AsRef<T> for FunctionTy<T> {
...
+
+            invokable_functions.insert(validator.common_keys.function);
+
+            invokable_functions.extend(validator.common_keys.resolver);
+        });
+
+        self.workflow_post_function
+            .iter()
+            .for_each(|post_function| {
+                invokable_functions.insert(post_function.common_keys.key);
```
jwong101, line 329: The `key` property isn't a function, it's the identifier for the module, so it shouldn't be in the set of invokable functions.

crates/forge_loader/src/manifest.rs
```diff
@@ -208,63 +282,102 @@ impl<T> AsRef<T> for FunctionTy<T> {
...
+
+                invokable_functions.insert(post_function.common_keys.function);
+
+                invokable_functions.extend(post_function.common_keys.resolver);
+            });
+
+        // get user invokable modules that have additional exposure endpoints.
+        // ie macros has config and export fields on top of resolver fields that are functions
+        self.macros.iter().for_each(|macros| {
+            invokable_functions.insert(macros.common_keys.key);
```
jwong101, line 339: The `key` property isn't a function.

crates/forge_loader/src/manifest.rs
```diff
@@ -86,11 +157,28 @@ pub struct ForgeModules<'a> {
...
+    pub ui_modifications: Vec<UiModificatons<'a>>,
+    #[serde(rename = "jira:workflowValidator", default, borrow)]
+    pub workflow_validator: Vec<WorkflowValidator<'a>>,
     #[serde(rename = "jira:workflowPostFunction", default, borrow)]
-    workflow_post_functions: Vec<WorkflowPostFunction<'a>>,
+    pub workflow_post_function: Vec<WorkflowPostFunction<'a>>,
+    // deserializing admin pages
+    #[serde(rename = "jira:adminPage", default, borrow)]
+    pub jira_admin: Vec<JiraAdminPage<'a>>,
     #[serde(flatten)]
```
jwong101, line 173: We can remove this property, since it's not being used anymore.

crates/forge_loader/src/manifest.rs
```diff
@@ -208,63 +282,102 @@ impl<T> AsRef<T> for FunctionTy<T> {
...
 
 impl<'a> ForgeModules<'a> {
-    pub fn into_analyzable_functions(
-        mut self,
-    ) -> impl Iterator<Item = FunctionTy<FunctionMod<'a>>> {
-        //FIXME: The logic here is incorrect, a function should be in the ignored function IIF it
-        //is not invoked by a user-invocable module.
-
+    // TODO: function returns iterator where each item is some specified type.
+    pub fn into_analyzable_functions(mut self) -> impl Iterator<Item = Entrypoint<'a>> {
```
jwong101, line 286: There's a possibility that whenever we add new invokable modules to the `ForgeModules` struct that we forget to add them to the logic here. If instead of accessing properties using `self.<prop>`, we do:

```rust
let Self { webtriggers, custom_field, etc } = self;
```

Then `rustc` will throw a compile error if we do not explicitly destructure it like above. That way, we won't forget to add them here. If a module is irrelevant to the logic below, we can destructure the property, but not bind it like so:

```rust
let Self {
    event_triggers: _, // NOTE: non-invokable module.
} = self;
```

Make sure that you don't use a catch-all pattern like `..` at the end, as that would defeat the purpose of the explicit destructuring.

jwong101 (changes requested):
Asides from the requested changes, could you also add the invokable [Jira Service Management](https://developer.atlassian.com/platform/forge/manifest-reference/modules/index-jsm/) and [https://developer.atlassian.com/platform/forge/manifest-reference/modules/index-bitbucket/](Bitbucket) modules as well?

crates/forge_loader/src/manifest.rs
```diff
@@ -25,24 +25,54 @@ pub struct FunctionMod<'a> {
...
 }
 
+// Abstracting away key, function, and resolver into a single struct for reuse whoo!
 #[derive(Default, Debug, Clone, PartialEq, Eq, Deserialize)]
-struct ModInfo<'a> {
+struct CommonKey<'a> {
     key: &'a str,
-    title: &'a str,
+    function: &'a str,
+    resolver: Option<&'a str>,
```
jwong101, line 33: This deserialization is incorrect, `resolver` should be an `Option<Resolver<'a>>` where `Resolver` is defined like so:

```rust
#[derive(Debug, Default, Clone, PartialEq, Eq, Deserialize)]
struct Resolver<'a> {
    function: Option<&'a str>,
}
```

See <https://developer.atlassian.com/platform/forge/build-a-custom-ui-app-in-confluence/#change-the-content-byline-item-title> as an example of how the `resolver` property should look.

Also, `function` should be an `Option<&'a str>`, since `Forge` modules do not have to specify a `function` if they specify a `resolver`. Otherwise, we'll error out whenever someone uses a module with a `resolver` and no `function`.

crates/forge_loader/src/manifest.rs
```diff
@@ -416,4 +529,159 @@ mod tests {
...
+                "name": "My App",
+                "id": "my-app"
+            },
+            "modules": {
+                "macro": [
+                {
+                    "key": "my-macro",
+                    "title": "My Macro",
+                    "function": "Catch-me-if-you-can0", 
+                    "resolver": "Catch-me-if-you-can1",
```
jwong101, line 547: See <https://github.com/atlassian-labs/FSRT/pull/31#discussion_r1451932098> for how the `resolver` property should look. Also, can you add test cases for when there's a `resolver`, but no outer `function` to make sure we handle the case I mentioned in <https://github.com/atlassian-labs/FSRT/pull/31#discussion_r1451932098>?

```suggestion
                    "resolver": {
                        "function": "Catch-me-if-you-can1"
                    },
```

crates/forge_loader/src/manifest.rs
```diff
@@ -25,24 +25,54 @@ pub struct FunctionMod<'a> {
     providers: Option<AuthProviders<'a>>,
 }
 
+// Abstracting away key, function, and resolver into a single struct for reuse whoo!
 #[derive(Default, Debug, Clone, PartialEq, Eq, Deserialize)]
-struct ModInfo<'a> {
+struct CommonKey<'a> {
```
jwong101, line 30: You might want to add a helper method for getting the functions of this struct, since you'll be calling it a lot. One approach would
be:

```rust
impl<'a> CommonKey<'a> {
    fn append_functions<I: Extend<Item = &'a str> + ?Sized>(&self, funcs: &mut I) {
        funcs.extend(self.key);
        if let Some(Resolver { function }) = self.resolver {
            funcs.extend(function);
        }
    }
}
```

So that you can use `module.common.append_functions(&mut invokable_functions);` for each of the invokable functions. That way, you're less likely to forget to potentially add either the `function` or `resolver.function` property.

crates/forge_loader/src/manifest.rs
aliner-wang: Done

crates/forge_loader/src/manifest.rs
aliner-wang: Done

crates/forge_loader/src/manifest.rs
```diff
@@ -59,23 +81,159 @@ enum Interval {
...
+    #[serde(flatten, borrow)]
+    common_keys: CommonKey<'a>,
+    edit: Option<&'a str>,
+}
+
+#[derive(Default, Debug, Clone, PartialEq, Eq, Deserialize)]
+struct IssueClass<'a> {
+    #[serde(flatten, borrow)]
+    common_keys: CommonKey<'a>,
+    dynamic_properties: JustFunc<'a>,
```
jwong101, line 199:
```suggestion
    #[serde(rename = "dynamicProperties")]
    dynamic_properties: Option<JustFunc<'a>>,
```
The `dynamic_properties` field is [optional](https://developer.atlassian.com/platform/forge/manifest-reference/modules/jira-issue-context/#properties) and in camelCase. You can also add `#[serde(rename_all = "camelCase")]` on top of `IssueClass` right below the `derive` instead of renaming the field explicitly.

crates/forge_loader/src/manifest.rs
```diff
@@ -59,23 +81,159 @@ enum Interval {
...
+    value: Option<&'a str>,
+    edit: Option<&'a str>,
+    context_config: Option<&'a str>,
+}
+
+#[derive(Default, Debug, Clone, PartialEq, Eq, Deserialize)]
+pub struct DashboardGadget<'a> {
+    #[serde(flatten, borrow)]
+    common_keys: CommonKey<'a>,
+    edit: Option<&'a str>,
```
jwong101, line 192:
```suggestion
    edit: Option<JustFunc<'a>>,
```
The edit property is a [struct](https://developer.atlassian.com/platform/forge/manifest-reference/modules/jira-dashboard-gadget/#properties) that may have a function property.

crates/forge_loader/src/manifest.rs
```diff
@@ -59,23 +81,159 @@ enum Interval {
...
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize)]
+pub struct CustomFieldType<'a> {
+    #[serde(flatten, borrow)]
+    common_keys: CommonKey<'a>,
+    value: Option<&'a str>,
+    edit: Option<&'a str>,
+    context_config: Option<&'a str>,
```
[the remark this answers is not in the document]
jwong101: Also, since `context_config` is camelCase in the manifest, you'll need to either add `#[serde(rename = "contextConfig")]` above it or add `#[serde(rename_all = "camelCase")]` at the top of the struct right below the derive.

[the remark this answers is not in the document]
jwong101: It might also be nice to leave a comment with a link to [this page](https://developer.atlassian.com/platform/forge/manifest-reference/modules/jira-custom-field-type/#search-suggestions) on top of the struct, explaining that `searchSuggestions` isn't user invokable, which is why we don't deserialize it. Rustdoc renders markdown, so you can use markdown embed linking in your comment if you want.

crates/forge_loader/src/manifest.rs
```diff
@@ -59,23 +81,159 @@ enum Interval {
...
+    resolver: Resolver<'a>,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize)]
+pub struct CustomField<'a> {
+    #[serde(flatten, borrow)]
+    common_keys: CommonKey<'a>,
+    value: Option<&'a str>,
+    search_suggestions: Option<&'a str>,
+    edit: Option<&'a str>,
```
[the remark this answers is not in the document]
jwong101: Also, `search_suggestions` isn't invokable according to the [docs](https://developer.atlassian.com/platform/forge/manifest-reference/modules/jira-custom-field/#search-suggestions), so we should probably skip deserializing it. You might want to leave a similar comment explaining why we don't deserialize it with a link to the docs.

crates/forge_loader/src/manifest.rs
```diff
@@ -59,23 +81,159 @@ enum Interval {
...
+#[derive(Default, Debug, Clone, PartialEq, Eq, Deserialize)]
+pub struct ComponentPage<'a> {
+    #[serde(flatten, borrow)]
+    common_keys: CommonKey<'a>,
+}
+#[derive(Default, Debug, Clone, PartialEq, Eq, Deserialize)]
+pub struct DataProvider<'a> {
+    #[serde(flatten, borrow)]
+    common_keys: CommonKey<'a>,
+    callback: JustFunc<'a>,
```
jwong101, line 124:
```suggestion
    callback: Option<JustFunc<'a>>,
```
This field is optional according to the [docs](https://developer.atlassian.com/platform/forge/manifest-reference/modules/compass-data-provider/#properties).

crates/forge_loader/src/manifest.rs
```diff
@@ -208,63 +401,239 @@ impl<T> AsRef<T> for FunctionTy<T> {
...
+        let Self {
+            mut webtriggers,
+            custom_field,
+            consumers: _,
+            functions,
+            event_triggers: _,
+            scheduled_triggers: _,
+            compass_admin_page,
+            component_page,
+            data_provider,
```
jwong101, line 416: After looking at the [data provider](https://developer.atlassian.com/platform/forge/manifest-reference/modules/compass-data-provider/) docs, I think this might not be invokable, since it can only receive [events](https://developer.atlassian.com/cloud/compass/integrations/create-a-data-provider-app/). There's no user context in the examples either, so we can probably treat this module as non invokable.

jwong101 (changes requested):
The modules are comprehensive now. Just a few more changes about how deserialization is done, and we should be good to go.

crates/forge_loader/src/manifest.rs
```diff
@@ -59,23 +81,159 @@ enum Interval {
...
+struct ContentAction<'a> {
+    #[serde(flatten, borrow)]
+    common_keys: CommonKey<'a>,
+}
+#[derive(Default, Debug, Clone, PartialEq, Eq, Deserialize)]
+struct ContentByLineItem<'a> {
+    #[serde(flatten, borrow)]
+    common_keys: CommonKey<'a>,
+    #[serde(borrow)]
+    dynamic_properties: JustFunc<'a>,
```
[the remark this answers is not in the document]
jwong101: Also, the `dynamic_properties` field should be deserialized as "dynamicProperties". You can do this by adding `#[serde(borrow, rename = "dynamicProperties")]` as a [field attribute](https://serde.rs/field-attrs.html#rename) above it, or by adding `#[serde(rename_all = "camelCase")]` as a [container attribute](https://serde.rs/container-attrs.html#rename_all) on top of the `ContentByLineItem` struct below the `derive`.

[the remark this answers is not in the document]
aliner-wang: added

crates/forge_loader/src/manifest.rs
```diff
@@ -59,23 +81,159 @@ enum Interval {
...
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize)]
+pub struct WorkflowPostFunction<'a> {
+    #[serde(flatten, borrow)]
+    common_keys: CommonKey<'a>,
+}
+// Jira Service Management Modules
+#[derive(Default, Debug, Clone, PartialEq, Eq, Deserialize)]
+struct AssetsImportType<'a> {
```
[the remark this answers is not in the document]
aliner-wang: added

crates/forge_loader/src/manifest.rs
```diff
@@ -208,65 +434,316 @@ impl<T> AsRef<T> for FunctionTy<T> {
...
+                return string == func.key;
+            }) || compass_admin_page.iter().any(|admin_function| {
+                let Some(string) = admin_function.function else {
+                    return false;
+                };
+                return string == func.key;
+            });
+
+            Ok::<_, Error>(Entrypoint {
+                function: FunctionRef::try_from(func)?,
```
aliner-wang, line 735: After destructuring and making all attributes refs unless necessary, Rust is throwing a trait bound error. And I need help to understand what's up. 

`error[E0277]: the trait bound `FunctionRef<'_>: From<&FunctionMod<'_>>` is not satisfied
   --> crates/forge_loader/src/manifest.rs:735:27
    |
735 |                 function: FunctionRef::try_from(func)?,
    |                           ^^^^^^^^^^^^^^^^^^^^^ the trait `From<&FunctionMod<'_>>` is not implemented for `FunctionRef<'_>`
    |
    = help: the trait `TryFrom<FunctionMod<'a>>` is implemented for `&'a FunctionRef<'a>`
    = note: required for `&FunctionMod<'_>` to implement `Into<FunctionRef<'_>>`
    = note: required for `FunctionRef<'_>` to implement `TryFrom<&FunctionMod<'_>>`
`

I'm kinda confused why it's mad that certain traits aren't implemented. But I'm guessing it has to do with how everything is a reference now so that values can be reused or smth. Anyways, this is the last blocker rn.

crates/forge_loader/src/manifest.rs
```diff
@@ -84,28 +242,110 @@ pub struct ForgeModules<'a> {
...
+    #[serde(rename = "jira:projectSettingsPage", default, borrow)]
+    project_settings_page: Vec<CommonKey<'a>>,
+    #[serde(rename = "jira:uiModificatons", default, borrow)]
+    pub ui_modifications: Vec<UiModificatons<'a>>,
+    #[serde(rename = "jira:workflowValidator", default, borrow)]
+    pub workflow_validator: Vec<WorkflowValidator<'a>>,
     #[serde(rename = "jira:workflowPostFunction", default, borrow)]
-    workflow_post_functions: Vec<WorkflowPostFunction<'a>>,
+    pub workflow_post_function: Vec<WorkflowPostFunction<'a>>,
+    // Jira Service Management Modules
```
aliner-wang, line 306: > Asides from the requested changes, could you also add the invokable [Jira Service Management](https://developer.atlassian.com/platform/forge/manifest-reference/modules/index-jsm/) and [https://developer.atlassian.com/platform/forge/manifest-reference/modules/index-bitbucket/](https://github.com/atlassian-labs/FSRT/pull/Bitbucket) modules as well?

Confusion moment here. So I added all the JSM Modules because they all have a function property and resolver property that can take a string referencing a function. Some might be as simple as returning a component but nonetheless. I'm a bit confused here, because I saw your later comment that said the modules looked comprehensive when JSM modules weren't included yet. 

Just wanted to make a note of it here. And clear this up. yee.

jwong101: Ah I missed those, yeah those should be added too.

crates/forge_loader/src/manifest.rs
jwong101: The `TryFrom` impl is backwards, you want to `impl<'a> TryFrom<&FunctionMod<'a>> for FunctionRef<'a>`. The `impl <'a> TryFrom<FunctionMod<'a>> for &'a FunctionRef<'a>` can only go from owned `FunctionMod<'a>` to borrowed `FunctionRef<'a>`.

To fix the failing test on line 885, just do `FunctionRef::try_from(&func_handler).unwrap()` instead of using `try_into`.

jwong101:
I think you still need to rebase.

PR #33, opened 2024-01-30
EAS-1974

aliner-wang:
Authorize logic wasn't terminating after inner if logic terminates when authorize call is picked up. Added return statement so that it does terminate.

### DIFF
--- a/crates/forge_analyzer/src/definitions.rs
+++ b/crates/forge_analyzer/src/definitions.rs
@@ -918,7 +918,6 @@ impl<'cx> FunctionAnalyzer<'cx> {
                 };
                 let first_arg = first_arg?;
                 let is_as_app = authn.first() == Some(&PropPath::MemberCall("asApp".into()));
-                let api_call = classify_api_call(first_arg);
                 match classify_api_call(first_arg) {
                     ApiCallKind::Unknown => {
                         if is_as_app {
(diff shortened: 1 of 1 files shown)


crates/forge_analyzer/src/definitions.rs
```diff
@@ -917,6 +918,7 @@ impl<'cx> FunctionAnalyzer<'cx> {
                 };
                 let first_arg = first_arg?;
                 let is_as_app = authn.first() == Some(&PropPath::MemberCall("asApp".into()));
+                let api_call = classify_api_call(first_arg);
```
jwong101, line 921: You can remove this, since this isn't being used anywhere.

crates/forge_analyzer/src/definitions.rs
```diff
@@ -952,20 +954,18 @@ impl<'cx> FunctionAnalyzer<'cx> {
...
             [PropPath::Def(def), ..] if self.res.is_imported_from(def, "@forge/api").is_some() => {
                 if let Some(ImportKind::Named(ref name)) =
                     self.res.is_imported_from(def, "@forge/api")
                 {
                     if *name == *"authorize" {
-                        Some(Intrinsic::Authorize(IntrinsicName::Other));
+                        return Some(Intrinsic::Authorize(IntrinsicName::Other));
                     } else {
                         return None;
                     }
```
jwong101, line 965:
```suggestion
                    }
```

jwong101 (changes requested):
Looks good, just two small nits.

crates/forge_analyzer/src/definitions.rs
aliner-wang: Done

crates/forge_analyzer/src/definitions.rs
aliner-wang: Done

Issue #29, opened 2023-11-30
Trying to run FSRT following EcoScanner vulnerability detection


sebD:
I ran this command:

```
fsrt ./my-project
```
Where my-project is the path towards the app I wish to test.

I expected this to happen:
The scanner evaluates the app and reports any issues

Instead, this happened:
The program exits with errors:

> thread 'main' panicked at crates/forge_analyzer/src/definitions.rs:359:18:
> expected class

`fsrt --version`: 
```
fsrt 0.1.0
```

<details><summary>Backtrace</summary>
<p>

```
thread 'main' panicked at crates/forge_analyzer/src/definitions.rs:359:18:
expected class

stack backtrace:
   0: rust_begin_unwind
             at /rustc/79e9716c980570bfd1f666e3b16ac583f0168962/library/std/src/panicking.rs:597:5
   1: core::panicking::panic_fmt
             at /rustc/79e9716c980570bfd1f666e3b16ac583f0168962/library/core/src/panicking.rs:72:14
   2: <forge_analyzer::definitions::Lowerer as swc_ecma_visit::Visit>::visit_var_declarator
   3: swc_ecma_visit::visit_decl
   4: <forge_analyzer::definitions::Lowerer as swc_ecma_visit::Visit>::visit_var_declarator
   5: swc_ecma_visit::visit_decl
   6: forge_analyzer::definitions::run_resolver
   7: fsrt::scan_directory
   8: fsrt::main
```

</p>
</details>

jeffryan-appfire:
I'm facing the same panic error when trying to run the scanner after seeing that Atlassian announced it'd start running the scanner on deployed marketplace applications. I'm simply running it locally to get a leg up on any warnings it may give. Here's the specific error I'm seeing:
```
[repos]$ RUST_BACKTRACE=1 fsrt ./jmcf-cloud
thread 'main' panicked at crates/forge_analyzer/src/definitions.rs:411:18:
expected class
stack backtrace:
   0: _rust_begin_unwind
   1: core::panicking::panic_fmt
   2: <forge_analyzer::definitions::Lowerer as swc_ecma_visit::Visit>::visit_var_declarator
   3: swc_ecma_visit::visit_decl
   4: forge_analyzer::definitions::run_resolver
   5: fsrt::scan_directory
   6: fsrt::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

[...]

PR #35, opened 2024-02-22
bugfix/handling-functions-in-objects

gersbach:
However, I noticed an issue where vulns in the previous vuln apps are not being detected. I tried against the main branch and it looks like there is an issue there.


gersbach:
> LGTM, the crash is gone with this commit. If you want you could include a dummy object:
> 
> ```js
> // Test case for issue #29
> const Something = {
>   someFunction() {}
> };
> ```
> 
> in one of the test apps as a test case.

sounds good. I'll add it and push

PR #36, opened 2024-02-26
feature/EAS-2049 FSRT - doesn't lower method definitions on object lits

gersbach:
This lowers functions within object literals.

### DIFF
--- a/crates/forge_analyzer/src/definitions.rs
+++ b/crates/forge_analyzer/src/definitions.rs
@@ -9,6 +9,7 @@ use forge_utils::{create_newtype, FxHashMap};
 use itertools::Itertools;
 use serde::{Deserialize, Serialize};
 use smallvec::{smallvec, SmallVec};
+use swc_core::ecma::ast::{MethodKind, PrivateMethod};
 use swc_core::{
     common::{Span, SyntaxContext, DUMMY_SP},
     ecma::{
@@ -200,7 +201,6 @@ pub fn run_resolver(
         };
         module.visit_with(&mut lowerer);
     }
-
     for (curr_mod, module) in modules.iter_enumerated() {
         let global_id = environment.get_or_reserve_global_scope(curr_mod);
         let mut global_collector = GlobalCollector {
@@ -1057,8 +1057,6 @@ impl<'cx> FunctionAnalyzer<'cx> {
             )));
         }
 
-        // defid_
-
         let obj = self.lower_expr(obj, None);
         let Operand::Var(mut var) = obj else {
             // FIXME: handle literals
@@ -1161,14 +1159,11 @@ impl<'cx> FunctionAnalyzer<'cx> {
         var
     }
 
-    fn resolve_class_prop(&self, obj: &Expr) -> Option<&Class> {
-        if let Expr::Ident(ident) = obj {
-            if let Some(defid) = self.res.sym_to_id(ident.to_id(), self.module) {
-                if let Some(potential_class) = self.body.class_instantiations.get(&defid) {
-                    if let Some(DefKind::Class(class_id)) = self.res.defs.defs.get(*potential_class)
-                    {
-                        return self.res.defs.classes.get(*class_id);
-                    }
+    fn resolve_class_prop(&self, ident: &Ident) -> Option<&Class> {
+        if let Some(defid) = self.res.sym_to_id(ident.to_id(), self.module) {
+            if let Some(potential_class) = self.body.class_instantiations.get(&defid) {
+                if let Some(DefKind::Class(class_id)) = self.res.defs.defs.get(*potential_class) {
+                    return self.res.defs.classes.get(*class_id);
                 }
             }
         }
@@ -1179,9 +1174,6 @@ impl<'cx> FunctionAnalyzer<'cx> {
         if let Expr::Ident(ident) = expr {
             if let Some(DefKind::Class(class)) = self.res.sym_to_def(ident.to_id(), self.module) {
                 let id = ident.to_id();
-                // let Some(def) = self.res.sym_to_id(id.clone(), self.module) else {
-                //     return Literal::Undef.into();
-                // };
                 if let Some((_, def_constructor)) = class
                     .pub_members
                     .iter()
@@ -1193,15 +1185,17 @@ impl<'cx> FunctionAnalyzer<'cx> {
             }
         }
         if let Expr::Member(MemberExpr { obj, prop, .. }) = expr {
-            if let Some(class) = self.resolve_class_prop(obj) {
-                if let MemberProp::Ident(ident) = prop {
-                    if let Some((_, def_constructor)) = class
-                        .pub_members
-                        .iter()
-                        .find(|(name, defid)| name == &ident.sym)
-                    {
-                        let var = self.body.get_or_insert_global(*def_constructor);
-                        return Operand::with_var(var);
+            if let Expr::Ident(ident) = &**obj {
+                if let Some(class) = self.resolve_class_prop(ident) {
+                    if let MemberProp::Ident(ident) = prop {
+                        if let Some((_, def_constructor)) = class
+                            .pub_members
+                            .iter()
+                            .find(|(name, defid)| name == &ident.sym)
+                        {
+                            let var = self.body.get_or_insert_global(*def_constructor);
+                            return Operand::with_var(var);
+                        }
                     }
                 }
             }
@@ -1254,12 +1248,12 @@ impl<'cx> FunctionAnalyzer<'cx> {
                 self.lower_expr(&arg.expr, Some(defid))
             })
             .collect();
+
         let callee = match callee {
             CalleeRef::Super => Operand::Var(Variable::SUPER),
             CalleeRef::Import => Operand::UNDEF,
             CalleeRef::Expr(expr) => self.get_operand_for_call(expr),
         };
-
         let first_arg = args.first().map(|expr| &*expr.expr);
         let intrinsic = self.as_intrinsic(&props, first_arg);
         let call = match intrinsic {
@@ -2205,7 +2199,6 @@ impl Visit for FunctionCollector<'_> {
     }
 
     fn visit_constructor(&mut self, n: &Constructor) {
-        //n.visit_children_with(self);
         if let Some(class_def) = self.curr_class {
             if let DefKind::Class(class) = self.res.clone().def_ref(class_def) {
                 if let Some((_, owner)) = &class
@@ -2252,7 +2245,6 @@ impl Visit for FunctionCollector<'_> {
     }
 
     fn visit_class_method(&mut self, n: &ClassMethod) {
-        //n.visit_children_with(self);
         if let Some(class_def) = self.curr_class {
             if let DefKind::Class(class) = self.res.clone().def_ref(class_def) {
                 if let Some((_, owner)) = &class.pub_members.iter().find(|(name, defid)| {
@@ -2466,6 +2458,24 @@ impl Visit for FunctionCollector<'_> {
                     }
                 }
             }
+            Some(Expr::Object(object_lit)) => {
+                object_lit.props.iter().for_each(|prop| {
+                    if let PropOrSpread::Prop(prop) = prop {
+                        if let Prop::Method(MethodProp { key, function }) = &**prop {
+                            if let Some(key_ident) = key.as_ident() {
+                                let former_parent = self.parent;
+                                self.parent = self.res.get_prop_from_ident(
+                                    self.module,
+                                    id.to_id(),
+                                    key_ident,
+                                );
+                                self.handle_function(&function, None);
+                                self.parent = former_parent;
+                            }
+                        }
+                    }
+                });
+            }
             _ => {}
         }
     }
@@ -2513,9 +2523,11 @@ impl Visit for FunctionCollector<'_> {
             let def = self
                 .res
                 .get_or_overwrite_sym(id, self.module, DefKind::Function(()));
+
+            let former_parent = self.parent;
             self.parent = Some(def);
             n.function.visit_with(self);
-            self.parent = None;
+            self.parent = former_parent;
         }
     }
 
@@ -2528,10 +2540,10 @@ impl Visit for FunctionCollector<'_> {
 impl FunctionCollector<'_> {
     fn handle_function(&mut self, n: &Function, owner: Option<DefId>) {
         let owner = self.parent.unwrap_or_else(|| {
-            if let Some(defid) = owner {
-                defid
-            } else if let Some(defid) = self.curr_function {
-                defid
+            if let Some(def_id) = owner {
+                def_id
+            } else if let Some(def_id) = self.curr_function {
+                def_id
             } else {
                 self.res
                     .add_anonymous("__UNKNOWN", AnonType::Closure, self.module)
@@ -2857,7 +2869,6 @@ impl Visit for Lowerer<'_> {
                                 /// TODO: track these
                                 Prop::Getter(_) | Prop::Setter(_) => {}
                                 Prop::Method(MethodProp { key, function }) => {
-                                    function.body.visit_with(self);
                                     if let Some(sym) = key.as_symbol() {
                                         let def_id_function = self.res.add_anonymous(
                                             sym.clone(),
@@ -2870,6 +2881,7 @@ impl Visit for Lowerer<'_> {
                                             .pub_members
                                             .push((sym, def_id_function));
                                     }
+                                    self.visit_function(function);
                                 }
                             },
                         }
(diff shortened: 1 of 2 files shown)


crates/forge_analyzer/src/definitions.rs
```diff
@@ -2466,6 +2478,26 @@ impl Visit for FunctionCollector<'_> {
...
+                            if let Some(key_ident) = key.as_ident() {
+                                if let Some(def_id_function) =
+                                    self.res
+                                        .get_prop_from_ident(self.module, id.clone(), key_ident)
+                                {
+                                    self.handle_function(
+                                        &*function,
+                                        Some(def_id_function.clone()),
+                                        true,
+                                    );
```
gersbach, line 2494: I am slightly confused about the use case for a parent here. Should this be assigned to the function within the object?

jwong101: Yeah, `parent` is essentially the definition that the created function will use. The `DefId` that's created for the `prop` should be the right one to assign it to.

crates/forge_analyzer/src/definitions.rs
```diff
@@ -1193,16 +1203,19 @@ impl<'cx> FunctionAnalyzer<'cx> {
             }
         }
         if let Expr::Member(MemberExpr { obj, prop, .. }) = expr {
-            if let Some(class) = self.resolve_class_prop(obj) {
-                if let MemberProp::Ident(ident) = prop {
-                    if let Some((_, def_constructor)) = class
-                        .pub_members
-                        .iter()
-                        .find(|(name, defid)| name == &ident.sym)
+            if let MemberProp::Ident(ident) = prop {
```
jwong101, line 1204: Looking at `resolve_class_prop` and `resolve_object_prop`, you should be able to accept `Ident` instead of `Expr` and just push the check to the caller. i.e. also check if `obj` is an `Expr::Ident`, and then pass the `ident` to both functions.

crates/forge_analyzer/src/definitions.rs
```diff
@@ -1193,16 +1203,19 @@ impl<'cx> FunctionAnalyzer<'cx> {
...
+                if let Some(class) = self.resolve_class_prop(obj) {
+                    if let Some((_, def_constructor)) =
+                        class.pub_members.iter().find(|(name, defid)| {
+                            return name == &ident.sym;
+                        })
                     {
                         let var = self.body.get_or_insert_global(*def_constructor);
                         return Operand::with_var(var);
                     }
+                } else if let Some(def_id_method) = self.resolve_object_prop(obj, prop) {
```
jwong101, line 1214: `resolve_obj_prop` should just take an `Ident` as the `prop`, since you already destructure the `MemberProp` here.

crates/forge_analyzer/src/definitions.rs
```diff
@@ -1193,16 +1203,19 @@ impl<'cx> FunctionAnalyzer<'cx> {
...
+                    let var = self.body.get_or_insert_global(def_id_method.clone());
```
jwong101, line 1215: This clone is unnecessary, since `def_id_method` is `Copy`.

crates/forge_analyzer/src/definitions.rs
```diff
@@ -2466,6 +2478,26 @@ impl Visit for FunctionCollector<'_> {
...
                 }
             }
+            Some(Expr::Object(object_lit)) => {
+                object_lit.props.iter().for_each(|prop| {
+                    if let PropOrSpread::Prop(prop) = prop {
+                        if let Prop::Method(MethodProp { key, function }) = &**prop {
+                            if let Some(key_ident) = key.as_ident() {
+                                if let Some(def_id_function) =
+                                    self.res
+                                        .get_prop_from_ident(self.module, id.clone(), key_ident)
```
jwong101, line 2486: This clone is unnecessary, since `DefId` is `Copy`. You can dereference it instead: `*id`.

crates/forge_analyzer/src/definitions.rs
```diff
@@ -2466,6 +2478,26 @@ impl Visit for FunctionCollector<'_> {
...
+                                {
+                                    self.handle_function(
+                                        &*function,
+                                        Some(def_id_function.clone()),
```
jwong101, line 2490: This clone is unnecessary, since `DefId` is `Copy`.

jwong101 (changes requested):
Haven't gotten around to testing it, but don't forget to add a Test case to see if we can properly follow method invocations.

crates/forge_analyzer/src/definitions.rs
```diff
@@ -3318,6 +3351,43 @@ impl Environment {
...
+    #[inline]
+    pub fn get_object(&self, mod_id: ModId, id: Id) -> Option<&Class> {
+        if let Some(def_id) = self.get_sym(id, mod_id) {
+            if let def_key = self.defs.defs[def_id] {
+                if let DefKey::Class(class) = def_key {
+                    return Some(&self.defs.classes[class]);
+                }
+            }
+        }
+        None
```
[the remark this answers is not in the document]
gersbach: I was looking at this page right now: https://doc.rust-lang.org/std/option/#the-question-mark-operator-. So, if the get_sym returns None, then the entire function returns none?

crates/forge_analyzer/src/definitions.rs
gersbach: Fixed. Thank you!

crates/forge_analyzer/src/definitions.rs
gersbach: Fixed. Thank you!

crates/forge_analyzer/src/definitions.rs
gersbach: I needed to add the to_id() function instead. The dereferencing didn't work.

crates/forge_analyzer/src/definitions.rs
gersbach: Fixed. I used the function in the res instead. Thank you

crates/forge_analyzer/src/definitions.rs
```diff
@@ -1175,6 +1176,15 @@ impl<'cx> FunctionAnalyzer<'cx> {
         None
     }
 
+    fn resolve_object_prop(&self, obj: &Expr, prop: &MemberProp) -> Option<&DefId> {
+        if let Expr::Ident(ident) = &*obj {
```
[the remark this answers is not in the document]
gersbach: Fixed. Thank you!

crates/forge_analyzer/src/definitions.rs
```diff
@@ -3318,6 +3351,43 @@ impl Environment {
...
+    pub fn get_prop_from_ident(&self, mod_id: ModId, id: Id, ident: &Ident) -> Option<&DefId> {
+        if let Some(class) = self.get_object(mod_id, id) {
+            if let Some((_, def_method)) = class.pub_members.iter().find(|(name, defid)| {
+                return name == &ident.sym;
+            }) {
+                return Some(def_method);
+            }
+        }
+        None
+    }
```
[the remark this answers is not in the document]
gersbach: Fixed. Thank you!

crates/forge_analyzer/src/definitions.rs
gersbach: I am now passing in the ident instead. I am slightly confused by pushing the check to the caller. The check to see if there is a property in the class? Sorry 😅

crates/forge_analyzer/src/definitions.rs
gersbach: Hope this works.

crates/forge_analyzer/src/definitions.rs
jwong101: Ah, the `get_prop_defid_from_object` method takes a `MemberProp`, however, it returns `None` if the `MemberProp` isn't an `Ident`. Furthermore, you already destructure the `MemberProp::Ident` variant before, the call, so you can just accept an `Ident` instead of `MemberProp`.

That way, if a caller already has an `Ident`, then you don't need to wrap it in a `MemberProp`, and it's more transparent on what's needed to get a property. This [article](https://matklad.github.io/2023/11/15/push-ifs-up-and-fors-down.html#Push-Ifs-Up) explains the concept pretty well.

test-apps/basic/src/index.tsx
```diff
@@ -14,6 +14,16 @@ let test_function = (word) => {
 }
 
 const App = () => {
+
+  let testObject = {
+    someFunction() {
+      let test_flag = "test_flag_within_function_within_object";
```
jwong101, line 20: Can you change this to perform an API call with `asApp()`? That way, we can check if we're also resolving in the interpreter/checkers.

crates/forge_analyzer/src/definitions.rs
```diff
@@ -38,6 +39,7 @@ use swc_core::{
         visit::{noop_visit_type, Visit, VisitWith},
     },
 };
+use time::format_description::modifier::Ignore;
```
jwong101, line 42: This import is unnecessary.

jwong101 (changes requested):

test-apps/basic/src/index.tsx
gersbach: Fixed. Thank you!

crates/forge_analyzer/src/definitions.rs
gersbach: Thank you. That article was very helpful. I do not require that check since it is already in the function. Resolved. Thanks again.

crates/forge_analyzer/src/definitions.rs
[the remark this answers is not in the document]
gersbach: fixed. Thanks!

test-apps/basic/src/index.tsx
```diff
@@ -14,6 +14,16 @@ let test_function = (word) => {
 }
 
 const App = () => {
+
+  let testObject = {
+    someFunction() {
+      return api.asApp().requestConfluence(route`/rest/api/3/test`);
```
jwong101, line 20: This test case is actually not being caught. We're getting baited by the `foo()` function call that also makes an `asApp()` call, which causes the `run` entrypoint to trigger the authorization vulnerability. If you remove the call to `asApp()` in `foo()`, then you'll notice that we aren't reporting any vulnerabilities here, even though we should.

You can either remove the call to `foo()` in the `App` function or you can create a separate module for testing object method calls.

For actually solving the problem of us not tracking method invocations, you'll probably have to look into the interpreter/checkers to see how we resolve calls. My gut tells me that we currently treat method calls as a no-op.

jwong101 (changes requested):

test-apps/basic/src/index.tsx
gersbach: Thank you very much for checking. This has been resolved. I think I introduced a bug when resolving merge conflicts 😅

jwong101 (changes requested):
Nice job! The test case passes now. However, we're hitting a bug, where the method calls are incorrectly being resolved to another object under the following conditions:

```javascript
  let foo = {
    someFunction() {},
  };
  let testObject = {
    someFunction() {
      const res = api.asApp().requestConfluence(route`/rest/api/3/test`);
      test_function('hi');
      return res;
    },
  };
  // this shouldn't trigger a vulnerability, but it currently is.
  foo.someFunction();
```

To be honest, I'm pretty sure this bug already existed before your PR, however, we didn't end up hitting it, because object methods weren't lowered in the first place. The bug is most likely because of how the interpreter/checkers resolve object method properties, but I could be wrong.

Regardless, we should probably also make the method calling more accurate in this PR as well due to the risk of reporting false positives. You could either:
1. Have the method invocation resolve to ambiguity if there are multiple possible methods and skip the function invocation. This would be the easiest way if you would like to defer more accurate method resolving to  another task/PR.
2. Figure out why the methods are being aliased and try to correctly assign them to their parent object.

Also, you can use `fsrt --dump-ir App ./test-apps/basic` to view the IR for a function. It doesn't help much in this case, since it doesn't show the `DefId` of the function on `foo.someFunction`, however, you might find some use for it. You could also tweak it to dump more relevant information if that helps you debug.

crates/forge_analyzer/src/definitions.rs
```diff
@@ -2168,7 +2162,7 @@ impl Visit for FunctionCollector<'_> {
                                 if let Some(defid) =
                                     self.res.get_sym(ident_property.to_id(), self.module)
                                 {
-                                    self.handle_function(&n.function, Some(defid));
+                                    self.handle_function(&*n.function, Some(defid));
```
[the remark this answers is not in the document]
gersbach: fixed. Thank you so much, Josh!

crates/forge_analyzer/src/definitions.rs
```diff
@@ -2870,6 +2877,7 @@ impl Visit for Lowerer<'_> {
                                             .pub_members
                                             .push((sym, def_id_function));
                                     }
+                                    self.visit_function(&**function);
```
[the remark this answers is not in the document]
gersbach: fixed.

crates/forge_analyzer/src/definitions.rs
```diff
@@ -2508,7 +2516,7 @@ impl Visit for FunctionCollector<'_> {
             self.res.resolver.defs[defid] = DefRes::Function(());
             self.res.overwrite_def(defid, DefRes::Function(()));
             self.curr_function = Some(defid);
-            self.handle_function(&n.function, Some(defid));
+            self.handle_function(&*n.function, Some(defid));
```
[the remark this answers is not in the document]
gersbach: Fixed. Thank you!

crates/forge_analyzer/src/definitions.rs
```diff
@@ -2466,6 +2458,22 @@ impl Visit for FunctionCollector<'_> {
...
+                object_lit.props.iter().for_each(|prop| {
+                    if let PropOrSpread::Prop(prop) = prop {
+                        if let Prop::Method(MethodProp { key, function }) = &**prop {
+                            if let Some(key_ident) = key.as_ident() {
+                                self.parent = self.res.get_prop_from_ident(
+                                    self.module,
+                                    id.to_id(),
+                                    key_ident,
+                                );
+                                self.handle_function(&*function, None);
```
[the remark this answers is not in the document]
gersbach: fixed. Thank you!

gersbach:
> Nice job! The test case passes now. However, we're hitting a bug, where the method calls are incorrectly being resolved to another object under the following conditions:
> 
> ```js
>   let foo = {
>     someFunction() {},
>   };
>   let testObject = {
>     someFunction() {
>       const res = api.asApp().requestConfluence(route`/rest/api/3/test`);
>       test_function('hi');
>       return res;
>     },
>   };
>   // this shouldn't trigger a vulnerability, but it currently is.
>   foo.someFunction();
> ```
> 
> To be honest, I'm pretty sure this bug already existed before your PR, however, we didn't end up hitting it, because object methods weren't lowered in the first place. The bug is most likely because of how the interpreter/checkers resolve object method properties, but I could be wrong.
> 
> Regardless, we should probably also make the method calling more accurate in this PR as well due to the risk of reporting false positives. You could either:
> 
> 1. Have the method invocation resolve to ambiguity if there are multiple possible methods and skip the function invocation. This would be the easiest way if you would like to defer more accurate method resolving to  another task/PR.
> 2. Figure out why the methods are being aliased and try to correctly assign them to their parent object.
> 
> Also, you can use `fsrt --dump-ir App ./test-apps/basic` to view the IR for a function. It doesn't help much in this case, since it doesn't show the `DefId` of the function on `foo.someFunction`, however, you might find some use for it. You could also tweak it to dump more relevant information if that helps you debug.

I am glad you caught this. This took me some time to debug, but I needed to reset the parent to none after the function after it was called. It was an error in this pull request and actually would have been an issue for all functions following regardless of name. Thank you again. Hope you have a good evening

crates/forge_analyzer/src/definitions.rs
```diff
@@ -2466,6 +2458,23 @@ impl Visit for FunctionCollector<'_> {
...
+                    if let PropOrSpread::Prop(prop) = prop {
+                        if let Prop::Method(MethodProp { key, function }) = &**prop {
+                            if let Some(key_ident) = key.as_ident() {
+                                self.parent = self.res.get_prop_from_ident(
+                                    self.module,
+                                    id.to_id(),
+                                    key_ident,
+                                );
+                                self.handle_function(&function, None);
+                                self.parent = None;
```
jwong101, line 2472: Shouldn't we be resetting the parent to the old parent? That way, the parent is more accurately tracked for nested objects. You can save the old parent before the `for_each` call and then restore the parent after it.

jwong101 (changes requested):
I tested your changes, and it looks good! This should be the last change required. Don't forget to squash your commits into 1 after you apply the change I suggested.

crates/forge_analyzer/src/definitions.rs
gersbach: Fixed. Thank you very much!

PR #37, opened 2024-03-06
chore/CAMPTASKS-442  - Convert Definition Analysis to use a ProjectionVec

gersbach:
This refactors code for definition analysis checking for headers within objects.

Also, clippy made me do some refactoring as well 🙂

### DIFF
--- a/crates/forge_analyzer/src/checkers.rs
+++ b/crates/forge_analyzer/src/checkers.rs
@@ -13,6 +13,7 @@ use std::{
 
 use tracing::{debug, info, warn};
 
+use crate::interp::ProjectionVec;
 use crate::{
     definitions::{Const, DefId, Environment, IntrinsicName, Value},
     interp::{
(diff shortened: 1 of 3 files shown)


gersbach:
> This looks good in terms of semantics. Out of curiosity, have you noticed if there were any slowdowns/increased memory with this PR? If not, I'll merge this PR, and we can deal with optimizing it later. If yes, one change you could try would be to have a two hashmaps, one for lvalues with nonempty projections, and one with just `DefId,VarId` pairs.

I did not benchmark the efficiency in this change. I didn't notice any degradations that were visible, but I was only running it on a small app. I can make the change for non empty projections and one with projections in this PR if you'd like.

jwong101:
I'll merge it for now, and we can check perf later.

jwong101:
Once you rebase it, I'll merge it in.

PR #38, opened 2024-03-10
feat/CAMPTASKS-395 : checking fetch calls for a const auth header

gersbach:
I'll squash this down once it is approved. Also {['Authorization']: blah} is another case that we should handle, but it might best to put that into a separate pull request.

### DIFF
--- a/crates/forge_analyzer/src/checkers.rs
+++ b/crates/forge_analyzer/src/checkers.rs
@@ -14,6 +14,7 @@ use std::{
 use tracing::{debug, info, warn};
 
 use crate::interp::ProjectionVec;
+use crate::utils::projvec_from_str;
 use crate::{
     definitions::{Const, DefId, Environment, IntrinsicName, Value},
     interp::{
@@ -21,13 +22,12 @@ use crate::{
         WithCallStack,
     },
     ir::{
-        Base, BasicBlock, BasicBlockId, BinOp, Inst, Intrinsic, Literal, Location, Operand,
-        Projection, Rvalue, VarId, VarKind, Variable,
+        Base, BasicBlock, BasicBlockId, Inst, Intrinsic, Literal, Location, Operand, Projection,
+        Rvalue, VarId, VarKind, Variable,
     },
     reporter::{IntoVuln, Reporter, Severity, Vulnerability},
     utils::{
-        add_const_to_val_vec, add_elements_to_intrinsic_struct, convert_operand_to_raw,
-        get_defid_from_varkind, get_prev_value, get_str_from_operand, return_value_from_string,
+        add_elements_to_intrinsic_struct, convert_operand_to_raw, get_defid_from_varkind,
         translate_request_type,
     },
     worklist::WorkList,
@@ -926,61 +926,56 @@ impl<'cx> Runner<'cx> for SecretChecker {
         state: &Self::State,
         operands: Option<SmallVec<[Operand; 4]>>,
     ) -> ControlFlow<(), Self::State> {
-        if let Intrinsic::SecretFunction(package_data) = intrinsic {
-            if let Some(operand) = operands
-                .unwrap_or_default()
-                .get((package_data.secret_position - 1) as usize)
-            {
-                match operand {
-                    Operand::Lit(lit) => {
-                        let vuln =
-                            SecretVuln::new(interp.callstack(), interp.env(), interp.entry());
-                        if let Literal::Str(_) = lit {
+        match intrinsic {
+            Intrinsic::SecretFunction(package_data) => {
+                if let Some(operand) = operands
+                    .unwrap_or_default()
+                    .get((package_data.secret_position - 1) as usize)
+                {
+                    {
+                        if interp.check_for_const(operand, def) {
+                            let vuln =
+                                SecretVuln::new(interp.callstack(), interp.env(), interp.entry());
                             info!("Found a vuln!");
                             self.vulns.push(vuln);
                         }
                     }
-                    Operand::Var(var) => {
-                        if let Base::Var(varid) = var.base {
-                            if let Some(value) =
-                                interp.get_value(def, varid, Some(var.projections.clone()))
-                            {
-                                match value {
-                                    Value::Const(_) | Value::Phi(_) => {
-                                        let vuln = SecretVuln::new(
-                                            interp.callstack(),
-                                            interp.env(),
-                                            interp.entry(),
-                                        );
-                                        info!("Found a vuln!");
-                                        self.vulns.push(vuln);
-                                    }
-                                    _ => {}
-                                }
-                            } else if let Some(VarKind::GlobalRef(def)) =
-                                interp.body().vars.get(varid)
-                            {
-                                if let Some(value) = interp.value_manager.defid_to_value.get(def) {
-                                    println!("value [] {value:?}");
-                                    match value {
-                                        Value::Const(_) | Value::Phi(_) => {
-                                            let vuln = SecretVuln::new(
-                                                interp.callstack(),
-                                                interp.env(),
-                                                interp.entry(),
-                                            );
-                                            info!("Found a vuln!");
-                                            self.vulns.push(vuln);
-                                        }
-                                        _ => {}
-                                    }
-                                }
-                            }
+                }
+            }
+            Intrinsic::Fetch => {
+                if let Some(Operand::Var(Variable {
+                    base: Base::Var(varid),
+                    ..
+                })) = operands.unwrap_or_default().get(1)
+                {
+                    let varid_argument =
+                        if let Some(Value::Object(varid)) = interp.get_value(def, *varid, None) {
+                            varid
+                        } else {
+                            varid
+                        };
+                    let headers_proj = projvec_from_str("headers");
+                    if let Some(Value::Object(varid)) =
+                        interp.get_value(def, *varid_argument, Some(headers_proj))
+                    {
+                        let auth_proj = projvec_from_str("Authorization");
+                        let aut_proj_lower = projvec_from_str("authorization");
+                        if let Some(Value::Const(_) | Value::Phi(_)) = interp
+                            .get_value(def, *varid, Some(auth_proj.clone()))
+                            .or_else(|| interp.get_value(def, *varid, Some(aut_proj_lower.clone())))
+                        {
+                            let vuln =
+                                SecretVuln::new(interp.callstack(), interp.env(), interp.entry());
+                            info!("Found a vuln!");
+                            self.vulns.push(vuln);
                         }
                     }
                 }
             }
+
+            _ => {}
         }
+
         ControlFlow::Continue(*state)
     }
 }
@@ -1005,13 +1000,7 @@ impl PermissionDataflow {
                 intrinsic_argument.second_arg = Some(
                     phi_val
                         .iter()
-                        .filter_map(|data| {
-                            if let Const::Literal(lit) = data {
-                                Some(lit.clone())
-                            } else {
-                                None
-                            }
-                        })
+                        .map(|Const::Literal(lit)| lit.clone())
                         .collect_vec(),
                 )
             }
@@ -1044,7 +1033,7 @@ impl<'cx> Dataflow<'cx> for PermissionDataflow {
         }
     }
 
-    fn transfer_intrinsic<C: crate::interp::Runner<'cx, State = Self::State>>(
+    fn transfer_intrinsic<C: Runner<'cx, State = Self::State>>(
         &mut self,
         interp: &mut Interp<'cx, C>,
         _def: DefId,
@@ -1290,7 +1279,7 @@ impl PermissionVuln {
     }
 }
 
-pub struct DefintionAnalysisRunner {
+pub struct DefinitionAnalysisRunner {
     pub needs_call: Vec<(DefId, Vec<Operand>, Vec<Value>)>,
 }
 
@@ -1335,9 +1324,9 @@ impl IntoVuln for PermissionVuln {
     }
 }
 
-impl<'cx> Runner<'cx> for DefintionAnalysisRunner {
+impl<'cx> Runner<'cx> for DefinitionAnalysisRunner {
     type State = PermissionTest;
-    type Dataflow = DefintionAnalysisRunner;
+    type Dataflow = DefinitionAnalysisRunner;
     const NAME: &'static str = "DefinitionAnalysis";
 
     fn visit_intrinsic(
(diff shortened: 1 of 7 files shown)


gersbach:
This currently doesn't handle the case insensitive case. Will fix and update.

gersbach:
> This currently doesn't handle the case insensitive case. Will fix and update.

fixed!

crates/forge_analyzer/src/checkers.rs
```diff
@@ -926,61 +927,60 @@ impl<'cx> Runner<'cx> for SecretChecker {
...
+                    ..
+                })) = operands.unwrap_or_default().get(1)
+                {
+                    let varid_argument =
+                        if let Some(Value::Object(varid)) = interp.get_value(def, *varid, None) {
+                            varid
+                        } else {
+                            varid
+                        };
+                    let headers_proj = projvec_from_str("Headers");
```
jwong101, line 958: The lowercase/uppercase normalization only applies to the actual headers themselves, not the headers property.

crates/forge_analyzer/src/checkers.rs
```diff
@@ -926,61 +927,60 @@ impl<'cx> Runner<'cx> for SecretChecker {
...
+                    let headers_proj_lower = projvec_from_str("headers");
+                    if let Some(Value::Object(varid)) = interp
+                        .get_value(def, *varid_argument, Some(headers_proj))
+                        .or_else(|| {
+                            interp.get_value(def, *varid_argument, Some(headers_proj_lower))
+                        })
+                    {
+                        let auth_proj = projvec_from_str("Authorization");
+                        let aut_proj_lower = projvec_from_str("authorization");
+                        if let Some(Value::Const(_) | Value::Phi(_)) = interp
```
jwong101, line 968: `Value::Phi` should only be classified as a vuln if each of the values it can alias are also constants. Otherwise, the following would be classified as vuln, which shouldnt:

```javascript
  let tmp = process.env.PROD_SECRET;

  // also test for if the `authorization` property contains some uppercase or
  // lowercase letters (they all get normalized to lowercase, so they all map to the same header)
  let secret = "test";
  if (process.env.PROD) {
    secret = tmp;
  }

  fetch("url", { headers: { authorization: secret } });

```

crates/forge_analyzer/src/checkers.rs
```diff
@@ -926,61 +927,60 @@ impl<'cx> Runner<'cx> for SecretChecker {
...
+                        };
+                    let headers_proj = projvec_from_str("Headers");
+                    let headers_proj_lower = projvec_from_str("headers");
+                    if let Some(Value::Object(varid)) = interp
+                        .get_value(def, *varid_argument, Some(headers_proj))
+                        .or_else(|| {
+                            interp.get_value(def, *varid_argument, Some(headers_proj_lower))
+                        })
+                    {
+                        let auth_proj = projvec_from_str("Authorization");
```
jwong101, line 966: Technically, any uppercase/lowercase combination of "authorization" works, however, these two are likely to be the only ones people use, so handling all cases can be a separate PR/task unless you want to do it in this one.

jwong101 (changes requested):
Looks good! I have two concerns, but I like the overall approach.

crates/forge_analyzer/src/checkers.rs
gersbach: fixed. Thank you!

crates/forge_analyzer/src/checkers.rs
gersbach: This has been fixed. Phi is supposed to be a vec of constants, but this will no longer be a false positive.

crates/forge_analyzer/src/checkers.rs
gersbach: should we loop through all of the characters, or do you think there is an alternative way to do this? Maybe make any vecs with properties authorization all lower case?

crates/forge_analyzer/src/interp.rs
```diff
@@ -676,9 +564,154 @@ impl<'cx, C: Runner<'cx>> Interp<'cx, C> {
...
+        } = lval
+        {
+            let (varid, projections) = self.get_farthest_obj(defid_block, varid, projections);
+            let rval_value = self.value_from_rval(defid_block, rvalue);
+            if let Some(existing_lval) = self
+                .get_value(defid_block, varid, Some(projections.clone()))
+                .cloned()
+            {
+                // if there is an existing value...
+                match (existing_lval, rval_value) {
```
jwong101, line 587: Don't forget to delete all subprojections when an object gets overwritten, otherwise the following test case will trigger:

```javascript
let obj = { headers: { authorization: 'secret' } };
obj = {};
fetch('test.com', obj);
```

jwong101 (changes requested):

crates/forge_analyzer/src/interp.rs
gersbach: fixed. Thank you for the help! BTreeMap worked amazing!

crates/forge_analyzer/src/interp.rs
```diff
@@ -676,9 +564,190 @@ impl<'cx, C: Runner<'cx>> Interp<'cx, C> {
...
+                            projections,
+                        ),
+                    (Value::Object(exist_var), Value::Object(new_var)) => {
+                        // store projection values that are transferred
+                        let mut projections_transferred = vec![];
+                        // tranfer all projection values from the new_var into the existing var
+                        for ((_, _, projections), value) in
+                            self.value_manager.varid_to_value.clone().range(
+                                (defid_block, new_var, ProjectionVec::new())
+                                    ..(defid_block, VarId(new_var.0 + 1), ProjectionVec::new()),
```
jwong101, line 627: There's technically a bug in the event that `new_var.0 = u32::MAX`, since this would overflow back to 0. You can either create separate ranges for this case:

```rust
let start = (defid_block, new_var, smallvec![]);
let query = match new_var.0.checked_add(1) {
    Some(end) => start..(defid_block, VarId(end), smallvec![]),
    None => start..
};
for placeholder in self.value_manager.var_id_to_value.range(query) {
// etc
```

See [u32::checked_add](https://doc.rust-lang.org/std/primitive.u32.html#method.checked_add) for more details. Also, instead of cloning the entire `value_manager.varid_to_value`, you can instead create a `Vec` of the cloned `(value, projections)` using [Iterator::collect](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.collect) or `collect_vec`, and loop over them and call `self.add_value_with_projection` afterwards.

jwong101 (changes requested):

crates/forge_analyzer/src/interp.rs
```diff
@@ -676,9 +564,190 @@ impl<'cx, C: Runner<'cx>> Interp<'cx, C> {
...
+                            projections_transferred.push(projections.clone());
+                            self.add_value_with_projection(
+                                defid_block,
+                                exist_var,
+                                value.clone(),
+                                projections.clone(),
+                            )
+                        }
+                        // clear any remaining values
+                        for ((_, _, _), value) in self
```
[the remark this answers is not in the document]
gersbach: Hmmm... I've spent some time trying to do this? Wouldn't we need to clone anyway since we are doing a query 2 times on a btreemap?

crates/forge_analyzer/src/interp.rs
gersbach: I've tried googling to get the end if the `checked_add(1)` is none. Is there anyway to do this? I couldn't see anyway to do this....

crates/forge_analyzer/src/interp.rs
```diff
@@ -676,9 +564,190 @@ impl<'cx, C: Runner<'cx>> Interp<'cx, C> {
...
+                    .zip(all_values.iter())
+                    .flat_map(|(a, b)| vec![a.clone(), b.clone()])
+                    .chain(quasis_as_values.iter().skip(all_values.len()).cloned())
+                    .chain(all_values.iter().skip(quasis_as_values.len()).cloned())
+                    .collect_vec();
+
+                return_combinations_phi(values_joined)
+            }
+            Rvalue::Bin(BinOp::Add, op1, op2) => {
+                let value_op1 = self.value_from_operand(defid_block, op1);
```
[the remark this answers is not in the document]
gersbach: fixed. Thank you! I am just going to exclude handling number tokens since none of our current scanners would require numbers....

crates/forge_analyzer/src/interp.rs
```diff
@@ -676,9 +564,190 @@ impl<'cx, C: Runner<'cx>> Interp<'cx, C> {
...
+
+    #[inline]
+    fn get_farthest_obj(
+        &self,
+        defid_block: DefId,
+        varid: VarId,
+        mut projections: ProjectionVec,
+    ) -> (VarId, ProjectionVec) {
+        let mut current_var_id = varid;
+        for (i, _) in projections.clone().iter().enumerate() {
```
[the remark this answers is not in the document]
gersbach: I needed to clone the projections here, not sure if there is a way around it, since we return the projections element.

crates/forge_analyzer/src/interp.rs
jwong101: You use a [one-sided range](https://doc.rust-lang.org/std/ops/struct.RangeFrom.html), since if `checked_add(1)` is `None`, then you just need the rest of the map. The one-sided range is `start..` in this case.

crates/forge_analyzer/src/interp.rs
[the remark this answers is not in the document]
jwong101: Not for the loop, since you're only using the index. Instead of doing `for (i, _) in projections.iter().enumerate()`, you can just loop over the indices directly, i.e. `0..projections.len()`.

crates/forge_analyzer/src/interp.rs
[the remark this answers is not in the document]
jwong101: Yeah, but you only need to clone the `new_var`'s `Value`s and `Projection`s. For instance:

```rust
    for ((_, _, _), value) in self.value_manager.varid_to_value.range_mut(
        (defid_block, exist_var, ProjectionVec::new())
            ..(defid_block, VarId(exist_var.0 + 1), ProjectionVec::new()),
    ) {
        *value = Value::Unknown
    }
    // tranfer all projection values from the new_var into the existing var
    let vals = self
        .value_manager
        .varid_to_value
        .range(
            (defid_block, new_var, ProjectionVec::new())
                ..(defid_block, VarId(new_var.0 + 1), ProjectionVec::new()),
        )
        .map(|((_, _, projections), value)| (projections.clone(), value.clone()))
        .collect_vec();
    for (projections, value) in vals {
        self.add_value_with_projection(defid_block, exist_var, value, projections)
    }
```

crates/forge_analyzer/src/interp.rs
gersbach: I am getting this compilation error. Do I need to find the max element of the var and the max element of the projection vec? Sorry for so many questions.... 

I was trying to find a way to solve it, but I was unsuccessful 😅 

```
25 |                           let query = match new_var.0.checked_add(1) {
    |  _____________________________________-
626 | |                             Some(end) => start..(defid_block, VarId(end), ProjectionVec::new()),
    | |                                          ------------------------------------------------------ this is found to be of type `std::ops::Range<(DefId, VarId, SmallVec<[Projection; 1]>)>`
627 | |                             None => start..
    | |                                     ^^^^^^^ expected `Range<(DefId, VarId, ...)>`, found `RangeFrom<(DefId, VarId, ...)>`
628 | |                         };
    | |_________________________- `match` arms have incompatible types
    |
```

crates/forge_analyzer/src/interp.rs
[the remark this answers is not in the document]
gersbach: fixed. Thank you!

crates/forge_analyzer/src/interp.rs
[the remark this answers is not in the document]
gersbach: When we add the values first, we call `self.get_farthest_obj()`. However, when we use the BTreeMap first, we don't check this instead assigning all of them to unknown. As a result, we cannot access the object via `self.get_farthest_object` since it is overwritten with unknown. But, I will implement the cloning only for the projections in the first one. Thank you!

crates/forge_analyzer/src/interp.rs
jwong101: Ah, my bad. You need to put the queries in the different arms like so: 

```rust
let start = (defid_block, new_var, smallvec![]);
let query = match new_var.0.checked_add(1) {
    Some(end) => self.value_manager.var_id_to_value.range(start..(defid_block, VarId(end), smallvec![])),
    None => self.value_manager.var_id_to_value.range(start..),
};
for placeholder in query {
// etc
```

I forgot that `x..y` and `x..` are two different types.

crates/forge_analyzer/src/interp.rs
```diff
@@ -676,9 +564,203 @@ impl<'cx, C: Runner<'cx>> Interp<'cx, C> {
...
         projections: ProjectionVec,
     ) {
+        let (varid, projections) = self.get_farthest_obj(defid_block, varid, projections);
         self.value_manager
             .insert_var_with_projection(defid_block, varid, projections, value);
     }
+
+    // this function takes in an operand checks for previous values and returns a value optional
+    #[inline]
+    pub fn add_value_to_definition(&mut self, defid_block: DefId, lval: Variable, rvalue: Rvalue) {
```
jwong101, line 574: Looks good! There are two edge cases that you need to handle though:

There's a bug where we set the underlying object's properties to `Unknown` if it's assigned to itself.

There's another bug where the alias isn't being updated if the property was an object. I believe this is because you're only getting the underlying object via `get_farthest_obj`, but not updating properties that may exist in the alias.

```javascript
let h = { headers: { authorization: 'foo' } };
h = h;
// Bug 1: Should trigger a vuln, currently doesn't
fetch('url', h)

// Bug 2: Shouldn't trigger a vuln, currently does
let c = h;
c.headers = {};
fetch('url', h);
```

Note that you have to be careful, since aliases are only updated in Javascript if a property is used, i.e:

```javascript
let c = h.headers;
// h.headers is still { authorization: 'foo' } after this assignment.
c = {};
```

jwong101 (changes requested):

crates/forge_analyzer/src/interp.rs
gersbach: fixed! Thank you!

PR #39, opened 2024-03-14
EAS-2074 : Make FSRT more performant with a definition analysis map without projections

gersbach:
Don't worry, not another 500 line PR 😅. This includes the commit of the previous PR. But, might as well open this.... 😄

### DIFF
--- a/crates/forge_analyzer/src/checkers.rs
+++ b/crates/forge_analyzer/src/checkers.rs
@@ -1481,7 +1481,6 @@ impl<'cx> Dataflow<'cx> for DefinitionAnalysisRunner {
                             }
                         }
                         /* should be expanded to include all cases ... */
-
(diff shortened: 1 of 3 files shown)


gersbach:
this has been fixed to include a hash map instead of the btree map when there are no projections. thank you!

gersbach:
Let me validate that this still works with the objects.

gersbach:
Ok, I validated it works with the test cases that have been looked at. 

I noticed a false negative case, where 

```

    let h = { headers: { authorization: 'foo' } };
    let c = h
   c.headers = { };
   c.headers.authorization = "test"

```

I'll look at this after I fix the other bug assigned to me.

jwong101 (approved):
Approved. Once you rebase this, I'll merge it in.

gersbach:
> Approved. Once you rebase this, I'll merge it in.

rebased! thank you!

PR #41, opened 2024-04-02
EAS-2088 : Create a Simple Test Framework

gersbach:


gersbach:
wait, i see it is failing the clippy check. let me fix that 🙂

gersbach:
> It's only the formatter that's failing at the moment. Make sure you run `cargo fmt --all`, which should resolve it.

nice! it worked. thank you

crates/fsrt/src/main.rs
```diff
@@ -184,33 +211,103 @@ fn collect_sourcefiles<P: AsRef<Path>>(root: P) -> impl Iterator<Item = PathBuf>
...
+}
 
-    let manifest = fs::read_to_string(&manifest_file).into_diagnostic()?;
-    let manifest: ForgeManifest<'_> = serde_yaml::from_str(&manifest).into_diagnostic()?;
-    let name = manifest.app.name.unwrap_or_default();
+#[tracing::instrument(level = "debug")]
+fn scan_directory<'a>(
+    dir: PathBuf,
+    opts: &Args,
+    test: bool,
```
jwong101, line 222: Instead of having the `scan_directory` function have a dummy project that it uses if it's being run as a test, it might be cleaner to isolate the IO/FS calls that occur within the function and create a trait that encapsulates that behavior. You can also wrap all the swc parser and loader structs within a common struct that will be embedded within the types that actually implement the trait.

crates/fsrt/src/main.rs
```diff
@@ -1,9 +1,14 @@
 #![allow(clippy::type_complexity)]
+
+mod test;
```
jwong101, line 3:
```suggestion
#[cfg(test)]
mod test;
```

Conditionally including the test module will speed up build times, as the test functionality will only get compiled if you're actually running `cargo test`.

crates/fsrt/src/main.rs
```diff
@@ -184,33 +211,103 @@ fn collect_sourcefiles<P: AsRef<Path>>(root: P) -> impl Iterator<Item = PathBuf>
...
+#[tracing::instrument(level = "debug")]
+fn scan_directory<'a>(
+    dir: PathBuf,
+    opts: &Args,
+    test: bool,
+    forge_test_proj: Vec<MockForgeProject>,
+) -> Option<Vec<Vulnerability>> {
+    let manifest_text;
+    let manifest = if test {
+        let mut forge_manifest_test = ForgeManifest::default();
```
jwong101, line 227: It might be better to have the manifest be created when instantiating the `MockForgeProject`s for each test case. You can add helper methods to make this less verbose if you'd like as well, as the easier it is to write test cases, the more likely we'll do it.

gersbach: Oh, yes. This sounds good. I was planning to do this in a different PR. I am trying to cut down on my diff size for faster turn around. But, yeah, I'd like the ability to add different files as well. I can change this 🙂

crates/fsrt/src/main.rs
gersbach: Fixed. Thank you!

crates/fsrt/src/main.rs
gersbach: I need some clarification here 😅. I am slightly confused, what would implement this trait.  Thank you in advance :) . sorry for the dumb question

gersbach: Fixed this, but lmk what you think. Thank you so much for the help again.

gersbach:
I'll rebase once it's approved :)

crates/fsrt/src/forge_project.rs
```diff
@@ -0,0 +1,165 @@
...
+#[derive(Debug)]
+pub(crate) struct ForgeProjectFromDir {
+    #[allow(dead_code)]
+    pub dir: PathBuf,
+    pub manifest_file_content: String,
+}
+
+#[derive(Debug, Clone)]
+pub(crate) struct MockForgeProject<'a> {
+    pub files_name_to_source: HashMap<String, String>,
```
jwong101, line 112:
```suggestion
    pub files_name_to_source: HashMap<PathBuf, String>,
```

Since, we might want to test non UTF-8 file names in the future.

crates/fsrt/src/forge_project.rs
```diff
@@ -0,0 +1,165 @@
...
+use swc_core::common::sync::Lrc;
+use swc_core::common::{FileName, Globals, Mark, SourceFile, SourceMap, GLOBALS};
+use swc_core::ecma::ast::EsVersion;
+use swc_core::ecma::parser::{parse_file_as_module, Syntax, TsConfig};
+use swc_core::ecma::transforms::base::resolver;
+use swc_core::ecma::visit::FoldWith;
+use tracing::debug;
+
+pub(crate) trait ForgeProjectTrait<'a> {
+    fn load_file(&self, path: PathBuf, _: Arc<SourceMap>) -> Arc<SourceFile>;
```
jwong101, line 17:
```suggestion
    fn load_file(&self, path: impl AsRef<Path>, _: Arc<SourceMap>) -> Arc<SourceFile>;
```

This way, we can pass `&str` and borrowed paths as the file name. Also, we are always borrowing `path`, so there's no need to have the caller create an owned version of the `Path`.

crates/fsrt/src/test.rs
```diff
@@ -0,0 +1,30 @@
+mod tests {
+    use crate::{scan_directory_test, MockForgeProject};
+    use forge_loader::manifest::{ForgeManifest, FunctionMod};
+    use std::collections::HashMap;
+
+    #[test]
+    fn test_simple() {
+        let forge_manifest = ForgeManifest::create_manifest_with_func_mod(FunctionMod {
```
jwong101, line 8: I think you forgot to add this function.

jwong101 (changes requested):

crates/fsrt/src/forge_project.rs
gersbach: fixed! thank you :)

crates/fsrt/src/test.rs
gersbach: fixed! thank you :)

crates/fsrt/src/forge_project.rs
gersbach: fixed! thank you :)

jwong101 (changes requested):

crates/fsrt/src/test.rs
```diff
@@ -0,0 +1,73 @@
+mod tests {
```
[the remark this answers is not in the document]
gersbach: fixed. thank you :)

crates/fsrt/src/main.rs
```diff
@@ -184,35 +109,25 @@ fn collect_sourcefiles<P: AsRef<Path>>(root: P) -> impl Iterator<Item = PathBuf>
...
-    manifest_file.push("manifest.yaml");
-    if !manifest_file.exists() {
-        manifest_file.set_extension("yml");
-    }
-    debug!(?manifest_file);
+pub(crate) fn scan_directory<'a>(
+    dir: PathBuf,
+    opts: &Args,
+    project: impl ForgeProjectTrait<'a> + std::fmt::Debug,
+) -> Option<Vec<Vulnerability>> {
```
jwong101, line 124:
```suggestion
) -> Result<Vec<Vulnerability>> {
```

That way, we can test if we get correct errors, and you don't have to keep calling `.ok()` on all of the returned errors to convert them to options.

crates/fsrt/src/main.rs
```diff
@@ -406,16 +325,18 @@ fn scan_directory(dir: PathBuf, function: Option<&str>, opts: &Args) -> Result<(
     if run_permission_checker && !perm_interp.permissions.is_empty() {
         reporter.add_vulnerabilities([PermissionVuln::new(perm_interp.permissions)]);
     }
-    let report = serde_json::to_string(&reporter.into_report()).into_diagnostic()?;
```
jwong101, line 409: You can probably return `Report` instead of the `Vec` of `Vulnerability`s here. That way, you can move the reporting logic out of the function, so that we don't unnecessarily print the report to stdout during testing. You can add a method like `pub fn into_vulnerabilities(self) -> Vec<Vulnerability>` to `Report`, so that the testing framework can take just the vulns.

crates/fsrt/src/forge_project.rs
```diff
@@ -0,0 +1,197 @@
...
+                .insert(PathBuf::from(file_name.trim()), file_source.to_string());
+        }
+
+        mock_forge_project
+    }
+}
+
+#[derive(Debug)]
+pub(crate) struct ForgeProjectFromDir {
+    #[allow(dead_code)]
```
jwong101, line 135: You can just delete this `dir` property if it isn't being used anywhere.

crates/fsrt/src/forge_project.rs
```diff
@@ -0,0 +1,197 @@
...
+
+    fn get_manifest(&self) -> ForgeManifest<'_> {
+        let out = serde_yaml::from_str(&self.manifest_file_content);
+        out.unwrap_or_default()
+    }
+}
+
+impl<'a> ForgeProjectTrait<'a> for MockForgeProject<'a> {
+    fn load_file(&self, p: impl AsRef<Path>, _: Arc<SourceMap>) -> Arc<SourceFile> {
+        let cm: Lrc<SourceMap> = Default::default();
```
jwong101, line 173: You should move the sourcemap to the `MockForgeProject` struct, so you don't have have to keep creating it here. Also, looking at this, you should probably call `self.cm.new_source_file(..)` in the test setup whenever adding a mock file, and store the `Arc<SourceFile>` in a `HashMap<PathBuf, Arc<SourceFile>>`. That way, whenever this method is called, you can just do `self.files_name_to_source.get(p.as_ref()).unwrap().clone()`.

crates/fsrt/src/main.rs
```diff
@@ -425,10 +346,28 @@ fn main() -> Result<()> {
         .with(EnvFilter::from_env("FORGE_LOG"))
         .init();
     let dirs = std::mem::take(&mut args.dirs);
-    let function = args.function.as_deref();
+
     for dir in dirs {
+        let mut manifest_file = dir.clone();
```
jwong101, line 351:
```suggestion
        let mut manifest_file = dir.join("manifest.yaml");
```

crates/fsrt/src/main.rs
```diff
@@ -425,10 +346,28 @@ fn main() -> Result<()> {
...
+        manifest_file.push("manifest.yaml");
+        if !manifest_file.exists() {
+            manifest_file.set_extension("yml");
+        }
+        debug!(?manifest_file);
+
+        let manifest_text = fs::read_to_string(&manifest_file)
+            .into_diagnostic()
```
jwong101, line 359:
```suggestion
            .into_diagnostic()?;
```

We should be returning an error if there's no `manifest_file`, since we can't scan.

crates/fsrt/src/test.rs
```diff
@@ -0,0 +1,71 @@
...
+        key: "main",
+        handler: "index.run",
+        providers: None,
+    });
+
+    let mut test_forge_project = MockForgeProject {
+        test_manifest: forge_manifest,
+        files_name_to_source: HashMap::new(),
+    };
+    test_forge_project.files_name_to_source.insert(
```
jwong101, line 17: You should probably have a function called `fn add_file(&mut self, p: impl Into<PathBuf>, source: impl ToString)` to `MockForgeProject` where you add the path and the source to the inner soucemap and HashMap. That way, you don't have to change all the tests, whenever we change the implementation of `MockForgeProject`.

jwong101 (changes requested):

crates/fsrt/src/main.rs
gersbach: Return the reporter here right? Report I think is just a string ...

jwong101: By report, I mean the `Report` type, which is the result of `reporter.into_report()`.

gersbach: Makes sense. Thank you 😃

gersbach: I was looking at the report variable. But, now it makes sense 🙂

crates/fsrt/src/main.rs
gersbach: Fixed. Thank you :)

crates/fsrt/src/test.rs
gersbach: Fixed. Thank you :)

crates/fsrt/src/main.rs
gersbach: Fixed. Thank you :)

crates/fsrt/src/forge_project.rs
gersbach: Fixed. Thank you :)

gersbach: But, please check to make sure I did this correctly

crates/fsrt/src/forge_project.rs
gersbach: This actually was being used. I have no idea why I did this....

crates/fsrt/src/main.rs
gersbach: Please check this, I think I may have fixed it. Thanks :)

crates/fsrt/src/main.rs
gersbach: Fixed. Thanks :)

crates/fsrt/src/main.rs
```diff
@@ -184,35 +109,25 @@ fn collect_sourcefiles<P: AsRef<Path>>(root: P) -> impl Iterator<Item = PathBuf>
         })
 }
 
+#[allow(dead_code)]
+pub(crate) fn scan_directory_test(
```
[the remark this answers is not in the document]
gersbach: moved. Thanks

crates/fsrt/src/forge_project.rs
```diff
@@ -0,0 +1,197 @@
...
+        }
+    }
+
+    fn get_manifest(&self) -> ForgeManifest<'_> {
+        let out = serde_yaml::from_str(&self.manifest_file_content);
+        out.unwrap_or_default()
+    }
+}
+
+impl<'a> ForgeProjectTrait<'a> for MockForgeProject<'a> {
```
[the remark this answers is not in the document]
gersbach: moved. Thanks

crates/fsrt/src/forge_project.rs
```diff
@@ -0,0 +1,197 @@
...
+}
+
+#[derive(Debug)]
+pub(crate) struct ForgeProjectFromDir {
+    #[allow(dead_code)]
+    pub dir: PathBuf,
+    pub manifest_file_content: String,
+}
+
+#[derive(Debug, Clone)]
```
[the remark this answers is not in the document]
gersbach: Moved. Thanks

crates/fsrt/src/forge_project.rs
```diff
@@ -0,0 +1,197 @@
...
+                def_id,
+                invokable: entrypoint.invokable,
+                webtrigger: entrypoint.web_trigger,
+                admin: entrypoint.admin,
+            })
+        }));
+    }
+}
+
+#[allow(dead_code)]
```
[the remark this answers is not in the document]
gersbach: fixed this 😅

crates/fsrt/src/main.rs
```diff
@@ -425,10 +344,40 @@ fn main() -> Result<()> {
...
     for dir in dirs {
+        let mut manifest_file = dir.join("manifest.yaml");
+        if !manifest_file.exists() {
+            manifest_file.set_extension("yml");
+        }
+        debug!(?manifest_file);
+
+        let manifest_text = fs::read_to_string(&manifest_file)
+            .into_diagnostic()?
+            .clone();
```
jwong101, line 357:
```suggestion
     ;
```

crates/fsrt/src/main.rs
```diff
@@ -425,10 +344,40 @@ fn main() -> Result<()> {
...
+
+        let forge_project_from_dir = ForgeProjectFromDir {
+            dir: dir.clone(),
+            manifest_file_content: manifest_text,
+        };
+
         debug!(?dir);
-        scan_directory(dir, function, &args)?;
+        let reporter_result = scan_directory(dir.clone(), &args, forge_project_from_dir);
```
jwong101, line 365:
```suggestion
        let reporter_result = scan_directory(dir, &args, forge_project_from_dir);
```

crates/fsrt/src/main.rs
```diff
@@ -425,10 +344,40 @@ fn main() -> Result<()> {
...
+        match reporter_result {
+            Result::Ok(report) => {
+                let report = serde_json::to_string(&report).into_diagnostic().unwrap();
+                debug!("On the debug layer: Writing Report");
+                match &args.out {
+                    Some(path) => {
+                        let _ = fs::write(path, report.clone()).into_diagnostic();
```
jwong101, line 372:
```suggestion
                        let _ = fs::write(path, report).into_diagnostic();
```

crates/fsrt/src/test.rs
```diff
@@ -0,0 +1,224 @@
...
+    #[allow(dead_code)]
+    fn get_secret_packages(&self) -> Vec<PackageData> {
+        vec![]
+    }
+
+    fn get_manifest(&self) -> ForgeManifest<'_> {
+        self.test_manifest.clone()
+    }
+
+    fn with_files_and_sourceroot<
```
jwong101, line 89: Is there a reason why this doesn't use the default implementation?

crates/fsrt/src/test.rs
```diff
@@ -0,0 +1,224 @@
...
+#[test]
+fn test_simple() {
+    let forge_manifest = ForgeManifest::create_manifest_with_func_mod(FunctionMod {
+        key: "main",
+        handler: "index.run",
+        providers: None,
+    });
+
+    let mut test_forge_project = MockForgeProject {
+        test_manifest: forge_manifest.clone(),
```
jwong101, line 164:
```suggestion
        test_manifest: forge_manifest,
```

crates/fsrt/src/test.rs
```diff
@@ -0,0 +1,224 @@
...
+#[test]
+fn test_secret_vuln() {
+    let forge_manifest = ForgeManifest::create_manifest_with_func_mod(FunctionMod {
+        key: "main",
+        handler: "index.run",
+        providers: None,
+    });
+
+    let mut test_forge_project = MockForgeProject {
+        test_manifest: forge_manifest.clone(),
```
jwong101, line 189:
```suggestion
        test_manifest: forge_manifest,
```

crates/fsrt/src/test.rs
```diff
@@ -0,0 +1,224 @@
...
+
+impl fmt::Debug for MockForgeProject<'_> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "Mock Forge Project {:?}", self.files_name_to_source)
+    }
+}
+
+#[allow(dead_code)]
+impl MockForgeProject<'_> {
+    pub fn files_from_string(string: String) -> Self {
```
jwong101, line 33:
```suggestion
    pub fn files_from_string(string: &str) -> Self {
```

Saves a clone, since we end up having to create new strings anyway + it's more convenient.

jwong101 (changes requested):

crates/fsrt/src/main.rs
gersbach: fixed. thank you :)

crates/fsrt/src/main.rs
gersbach: fixed. thank you :)

crates/fsrt/src/test.rs
gersbach: fixed. thank you :)

crates/fsrt/src/test.rs
gersbach: fixed. thank you :)

crates/fsrt/src/test.rs
gersbach: fixed. thank you :)

crates/fsrt/src/main.rs
gersbach: fixed. thank you .... needed to remove this form the print statemtn below this if i wanted to remove the clone :)

crates/fsrt/src/test.rs
gersbach: nope. I removed it

crates/fsrt/src/main.rs
```diff
@@ -1,27 +1,24 @@
 #![allow(clippy::type_complexity)]
+
+mod forge_project;
+#[cfg(test)]
+#[allow(dead_code)]
```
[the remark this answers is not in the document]
gersbach: i removed it entirely

crates/forge_loader/src/manifest.rs
```diff
@@ -414,6 +414,17 @@ pub struct ForgeManifest<'a> {
     pub permissions: Perms<'a>,
 }
 
+impl<'a> ForgeManifest<'a> {
+    pub fn create_manifest_with_func_mod(function_mod: FunctionMod<'a>) -> Self {
+        let mut forge_manifest_test = ForgeManifest::default();
+        forge_manifest_test
+            .modules
+            .functions
+            .push(function_mod.clone());
```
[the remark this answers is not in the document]
gersbach: fixed.

crates/forge_analyzer/src/reporter.rs
```diff
@@ -28,7 +28,7 @@ pub trait IntoVuln {
 
 #[derive(Debug, Clone, PartialEq, Eq, Serialize)]
 pub struct Report {
-    vulns: Vec<Vulnerability>,
+    pub vulns: Vec<Vulnerability>,
```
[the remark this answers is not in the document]
gersbach: fixed.

PR #43, opened 2024-05-31
feature/bugfix/EAS-2171 : Write Test Cases for Object Literals

gersbach:
woohoo ! some bug fixes and testing for object literals :)

### DIFF
--- a/crates/forge_analyzer/src/interp.rs
+++ b/crates/forge_analyzer/src/interp.rs
@@ -425,6 +425,25 @@ impl ValueManager {
                 .insert((def_id_func, var_id, projection_vec), value);
         }
     }
+
+    pub fn get_var_with_projection(
+        &self,
+        def_id_func: DefId,
+        var_id: VarId,
+        projection_vec: ProjectionVec,
+    ) -> Option<&Value> {
+        if let Some(value) =
+            self.varid_to_value_with_proj
+                .get(&(def_id_func, var_id, projection_vec.clone()))
+        {
+            Some(value)
+        } else if let Some(Value::Object(var_id)) = self.varid_to_value.get(&(def_id_func, var_id))
+        {
+            self.get_var_with_projection(def_id_func, *var_id, projection_vec)
+        } else {
+            None
+        }
+    }
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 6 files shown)


crates/forge_analyzer/src/interp.rs
```diff
@@ -425,6 +425,25 @@ impl ValueManager {
...
+
+    pub fn get_var_with_projection(
+        &self,
+        def_id_func: DefId,
+        var_id: VarId,
+        projection_vec: ProjectionVec,
+    ) -> Option<&Value> {
+        if let Some(value) =
+            self.varid_to_value_with_proj
+                .get(&(def_id_func, var_id, projection_vec.clone()))
```
gersbach, line 437: is there any way to not clone this here ...

crates/fsrt/src/test.rs
```diff
@@ -55,37 +66,50 @@ impl fmt::Debug for MockForgeProject<'_> {
...
-        });
+impl<'a> MockForgeProject<'a> {
+    pub fn files_from_string(string: &'a str) -> Self {
+        let different_files = string
+            .split("//")
+            .map(|f| f.replace("//", "").trim().to_string())
+            .filter(|file| !file.is_empty());
+
+        let manifest = if string.contains("manifest.yaml") || string.contains("manifest.yml") {
+            let manifest_spelling = if string.contains("manifest.yaml") {
```
gersbach, line 77: this code is small brain. Any way to write this better?

crates/forge_analyzer/src/interp.rs
jwong101: Unfortunately not, since `std`'s trait for heterogeneous lookups are kind of scuffed. If you want you can add the [hashbrown](https://github.com/rust-lang/hashbrown) crate(this is literally `std::collections::HashMap`, but with more features that they don't want to stabilize yet. You'd use their [get](https://docs.rs/hashbrown/latest/hashbrown/struct.HashMap.html#method.get) method which takes something that implements [Equivalent](https://docs.rs/hashbrown/latest/hashbrown/trait.Equivalent.html). So you'd do something like

```rust
#[derive(PartialEq, Eq, Clone, Copy, Hash)]
struct ProjLookup<'a>(DefId, VarId, &'a [Value]);

impl Equivalent<&(DefId, VarId, SmallVec<Value>)> for ProjLookup<'_> {
    fn equivalent(&self, key: &(DefId, VarId, SmallVec<Value>) -> bool {
        self.0 == key.0 && self.1 == key.1 && self.2 == key.2[..]
    }
}

if let Some(value) = self.varid_to_value_with_proj.get(&ProjLookup(def_id_func, var_id, &projection_vec[..])
```

gersbach: Will take a look at the hashbrown documentation. However, I think we should eventually refractor them all this way, so I will leave this out of this PR.

crates/fsrt/src/test.rs
gersbach: fixed! Thanks as always, @jwong101

gersbach:
Should be ready for merge

PR #47, opened 2024-07-19
BUG-2199: Fix for the FSRT Multiple Definitions for Same Variable Bug

swathiatgit:
FSRT should be in “static single assignment form”, i.e. there should be exactly one definition of each variable, as this simplifies later analysis. This PR fixes this bug, and ensures that every variable has one definition.

More on the bug: Take this code example below:

```
function foo(a) {
  let b = a - 20;
  let c = b + 1;
  b = 10;
  return b * 10;
}
```
The above function currently gives this IR:

```
Variables:
%0: return value
%1: local definition of b
%2: local definition of c
%3: local definition of b
%4: local definition of b
%5: global ref to a
%6: temporary
%7: global ref to b
%8: temporary
%9: global ref to c
%10: temporary
bb0:
    %6 = %5 - 20
    %7 = %6 
    %8 = %7 + 1
    %9 = %8
    %7 = 10 // second assignment of %7
    _ = 10
    %10 = %7 * 10
    %0 = %10
    return
```
Note that in the current output of our IR that every “use” of the variable %7 only needs to consider the definition of “%7 = 10”. Thus, we can replace the assignment “%7 = 10” with “%11 = 10” and all following uses of “%7” with the new variable “%11”.

### DIFF
--- a/crates/forge_analyzer/src/definitions.rs
+++ b/crates/forge_analyzer/src/definitions.rs
@@ -1,13 +1,10 @@
 #![allow(dead_code, unused)]
 
-use std::hash::Hash;
 use std::{borrow::Borrow, fmt, mem};
 
 use crate::utils::{calls_method, eq_prop_name};
 use forge_file_resolver::{FileResolver, ForgeResolver};
 use forge_utils::{create_newtype, FxHashMap};
-use std::collections::{HashMap, HashSet};
-use swc_core::common::pass::define;
 
 use itertools::Itertools;
 use serde::{Deserialize, Serialize};
@@ -187,6 +184,7 @@ pub fn run_resolver(
         module.visit_with(&mut import_collector);
     }
 
+    // check for required after Definitions pass
     // This loop runs through the different import modules and corresponding definitions.
     let defs = Definitions::new(
         environment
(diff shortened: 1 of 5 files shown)


crates/forge_analyzer/src/definitions.rs
```diff
@@ -1464,7 +1467,7 @@ impl<'cx> FunctionAnalyzer<'cx> {
                                     let id = ident.to_id();
                                     let new_def =
                                         self.res.get_or_insert_sym(id.clone(), self.module);
-                                    let var_id = self.body.get_or_insert_global(new_def);
+                                    let var_id = self.body.insert_global(new_def);
```
gersbach, line 1470: @jwong101 @jwong33 i know we haven't been strict about LocalDef vs GlobalRef previously... do we want to start?

gersbach:
Very clean code 🙂

crates/forge_analyzer/src/definitions.rs
jwong101: Yeah, we shouldn't be creating new global variables/var ids, since this will create two separate variables:

```javascript
const b = 3;
const a = { b };
```

```
%b = 3
%a.b = %newb
```

jwong101 (changes requested):
The main problem with doing this during lowering is that you need to keep track of which uses of a symbol correspond to the new definition that you created. Right now it's implicitly the last one, which doesn't work for control flow across basic blocks.

Also, new uses of an identifier shouldn't create another VarId, rather it should refer to the last VarId that was assigned.

gersbach:
> The main problem with doing this during lowering is that you need to keep track of which uses of a symbol correspond to the new definition that you created. Right now it's implicitly the last one, which doesn't work for control flow across basic blocks.
> 
> Also, new uses of an identifier shouldn't create another VarId, rather it should refer to the last VarId that was assigned.
> 
> Regardless, nice work so far! Let me know if you have any questions.

@jwong101, I thought this bug was for within basic blocks though? Shouldn't we wait until the dominator tree is merged in and we can track between basic blocks then?

swathiatgit:
Thank you for the feedback @jwong101, taking a look at this right now. I believe the new uses of an identifier case should be fairly simple to resolve - I'm investigating it now. Regarding the issue with lowering and needing to track which uses of a symbol correspond to the new definition created, let me ensure that it's explicitly defined and not implicitly.

Ideally, if it's explicitly defined, I think that should translate fairly smoothly to incorporating the dominator tree work as well. I'll touch base with you and @gersbach regarding this soon when I've done more work to research it. Thank you!

jwong101 (changes requested):
Added comments where extraneous var ids are created.

While you can go with this approach, it will make the renaming trickier in later stages. Mainly because you now have to do another lookup to check the "true" alias of a var ID on every use to do the correction. I.e.:

```javascript
// mapping
// %0 -> a
// %1 -> b
// %2 -> a
// %3 -> c
// %4 -> d

let a = 1; // %0 = 1
if (true) {
  let b = a // %1 = %0
  a = 3; // %2 = 3
  let c = a; // %3 = %2
  return;
}
let d = a; // %4 = %2 ((assuming we visit the if statement bb first)
```

While you could solve this by checking the mapping, it's a lot easier to just treat the raw var IDs as the potential aliases.

Though one thing you could do in the lowerering code is treating global assignments differently(basically variables that aren't defined in the function body). Those can't be in SSA form, since the assignment semantics are different. (they should have a separate IR statement tbh)

PR #48, opened 2024-07-26
EAS-2209 Adding Dominators

jxuatlas:


crates/forge_analyzer/src/ir.rs
```diff
@@ -392,13 +416,185 @@ impl Body {
...
+    pub(crate) fn dominance_frontier(&self, b: BasicBlockId) -> impl Iterator<Item = BasicBlockId> + '_ {
+        let idom = &self.dominator_tree().idom;
+        let mut frontiers: Vec<Vec<BasicBlockId>> = Vec::new();
+        for _ in 0..self.blocks.len() {
+            frontiers.push(Vec::new());
+        }
+
+        // algorithm from https://en.wikipedia.org/wiki/Static_single-assignment_form#Computing_minimal_SSA_using_dominance_frontiers
+        // which references: https://www.cs.tufts.edu/comp/150FP/archive/keith-cooper/dom14.pdf
+        for (id, block) in self.iter_blocks_enumerated() {
```
aliner-wang, line 567: replace "block" with _ since you're not using "block" in the loop.

crates/fsrt/src/main.rs
```diff
@@ -171,6 +175,12 @@ pub(crate) fn scan_directory<'a>(
         std::process::exit(0);
     }
 
+    if let Some(func) = opts.dump_dt.as_ref() {
+        println!("My CLI thing WORKS");
```
aliner-wang, line 179: This was mb. Can you delete this line? ty <3

crates/forge_analyzer/src/ir.rs
jxuatlas: noted - thank you @aliner-wang! I think I have some other places with the same issue, will make those changes as well

crates/fsrt/src/main.rs
```diff
@@ -343,6 +353,7 @@ pub(crate) fn scan_directory<'a>(
 
 fn main() -> Result<()> {
     let mut args = Args::parse();
+    println!("What is args {:?}", args);
```
aliner-wang, line 356: Also delete this line pls or keep it (up to you) <3

PR #50, opened 2024-08-06
BUG-2199 + EAS-2222: Fix for the FSRT Multiple Definitions for Same Variable Bug

swathiatgit:
FSRT should be in “static single assignment form”, i.e. there should be exactly one definition of each variable, as this simplifies later analysis. This PR fixes this bug with mutable iteration over the created environment, and ensures that every variable has one definition.

More on the bug: Take this code example below:

```
function foo(a) {
  let b = a - 20;
  let c = b + 1;
  b = 10;
  return b * 10;
}
```
The above function currently gives this IR:

```
Variables:
%0: return value
%1: local definition of b
%2: local definition of c
%3: local definition of b
%4: local definition of b
%5: global ref to a
%6: temporary
%7: global ref to b
%8: temporary
%9: global ref to c
%10: temporary
bb0:
    %6 = %5 - 20
    %7 = %6 
    %8 = %7 + 1
    %9 = %8
    %7 = 10 // second assignment of %7
    _ = 10
    %10 = %7 * 10
    %0 = %10
    return
```
Note that in the current output of our IR that every “use” of the variable %7 only needs to consider the definition of “%7 = 10”. Thus, we can replace the assignment “%7 = 10” with “%11 = 10” and all following uses of “%7” with the new variable “%11”.


crates/forge_analyzer/src/definitions.rs
```diff
@@ -226,9 +233,96 @@ pub fn run_resolver(
...
+                        update_rvalue(rvalue, &vars_map);
+                        let var_id = variable.as_var_id().unwrap();
+                        let (var_kind, updated_var_id, global_exists) =
+                            vars_map.get_mut(&var_id).unwrap();
+                        if !*global_exists {
+                            *updated_var_id = Some(var_id);
+                            *global_exists = true;
+                        } else {
+                            let new_var_id = body.vars.push_and_get_key(*var_kind);
+                            variable.base = Base::Var(new_var_id);
```
jwong101, line 258: Don't forget to check that the projection's array is empty, since

```javascript
let a = { foo: 1 };
a.bar = 2;
a.foo = 3;
```

`a` should resolve to the same def id.

crates/forge_analyzer/src/definitions.rs
```diff
@@ -226,9 +233,96 @@ pub fn run_resolver(
...
+// Input: rvalue and hashmap of VarIds that have been updated in the SSA Form Loop.
+// Output: new rvalue that has the newest VarId in its operands.
+pub fn update_rvalue(
+    rvalue: &mut Rvalue,
+    vars_map: &HashMap<VarId, (VarKind, Option<VarId>, bool)>,
+) {
+    match rvalue {
+        Rvalue::Unary(unop, Operand::Var(variable)) => {
+            let op_var_id = variable.as_var_id().unwrap();
+            if (vars_map.contains_key(&op_var_id)) && (vars_map.get(&op_var_id).unwrap().2) {
```
jwong101, line 282:
```suggestion
            if let Some((_, updated_var_id, true)) = vars_map.get(&op_var_id) {
```

You can avoid the `contains_key` call by directly pattern matching, since [HashMap::get](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.get) already does the check and returns `None` if the key isn't present. By pattern matching, you can also combine the check for checking if the global is present.

jwong101 (changes requested):

crates/forge_analyzer/src/definitions.rs
```diff
@@ -226,9 +233,99 @@ pub fn run_resolver(
...
+            let var_kind = *body.vars.get(var_id).unwrap();
+            vars_map.insert(var_id, (var_kind, None, false));
+        }
+        // Updates the instruction set with creation of new global reference variables if necessary.
+        for (block_id, block) in body.blocks.iter_mut_enumerated() {
+            for inst in block.iter_insts_mut() {
+                match inst {
+                    Inst::Assign(variable, rvalue) => {
+                        update_rvalue(rvalue, &vars_map);
+                        let var_id = variable.as_var_id().unwrap();
```
jwong101, line 250: `variable.as_var_id()` is `None` when `variable` is `super` or `this`. The code should skip over the renaming instead of unwrapping the option

jwong101 (changes requested):

crates/forge_analyzer/src/definitions.rs
```diff
@@ -226,9 +233,92 @@ pub fn run_resolver(
         module.visit_with(&mut collector);
     }
 
+    // This loop iterates through env's bodies and recreates an updated set of bodies to satisfy SSA form of IR dump.
+    let mut vars_map: HashMap<VarId, (VarKind, Option<VarId>, bool)> = HashMap::new();
+    for (func_id, body) in environment.bodies_mut().enumerate() {
+        // Builds the variables map to track existing all variables, store their VarId, and check if they have been assigned once (by SSA).
+        for var_id in body.vars.keys() {
```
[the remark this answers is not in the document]
jwong101: Actually wait, do we need `Option<VarId>` if we can just insert the `VarId` into the map if its only been assigned? We can also probably use a separate `TiVec` of bools, since that has to be present for every `VarId`.

You should be able to get the `VarKind` by directly indexing `body.vars` as well.

jwong101 (approved):
Lgtm. Once you rebase/cleanup your commits I'll merge it in.

PR #51, opened 2024-08-13
2225

jxuatlas:


crates/forge_analyzer/src/definitions.rs
```diff
@@ -1695,17 +1703,20 @@ impl<'cx> FunctionAnalyzer<'cx> {
...
                 let curr = self.block;
-                let rest = self.body.new_block();
-                let cons_block = self.body.new_block();
-                let alt_block = self.body.new_block();
+                let [temp1, temp2, temp3] = self.body.new_blocks();
+                let rest = self.body.new_blockbuilder();
+                let cons_block = self.body.new_blockbuilder();
+                let alt_block = self.body.new_blockbuilder();
                 self.set_curr_terminator(Terminator::If {
+                    // TODO: COME BACK TO?
```
aliner-wang, line 1711: wdym come back to?

crates/forge_analyzer/src/definitions.rs
```diff
@@ -1846,17 +1875,29 @@ impl<'cx> FunctionAnalyzer<'cx> {
             Stmt::Labeled(LabeledStmt { label, body, .. }) => {
                 self.lower_stmt(body);
             }
+            // TODO: Lower Break and Continue
```
aliner-wang, line 1878: Is there a backlog ticket for this? If not, can you add one? :3

crates/forge_analyzer/src/definitions.rs
jxuatlas: Oh wait this was a note I made to myself while coding, thought I got rid of it in the latest commit but think I forgot to delete, will take out!

PR #54, opened 2024-11-12
EAS-2421 : Prevent running the scope scanner on Forge apps with remote auth

gersbach:


crates/forge_loader/src/manifest.rs
```diff
@@ -383,6 +383,18 @@ pub struct Perms<'a> {
     content: Content<'a>,
 }
 
+#[derive(Default, Debug, Clone, PartialEq, Eq, Deserialize)]
+pub struct Remotes<'a> {
+    #[serde(default, borrow)]
+    pub auth: &'a str,
```
jwong101, line 389: `auth` should be an optional object not a string as per the [manifest spec](https://developer.atlassian.com/platform/forge/manifest-reference/remotes/#remotes). This will fail to deserialize if someone does add a valid object.

You can use [serde_yaml::Value](https://docs.rs/serde_yaml/latest/serde_yaml/enum.Value.html) with `#[serde(default)]` and use https://docs.rs/serde_yaml/latest/serde_yaml/enum.Value.html#method.is_mapping to check if there's a valid mapping present.

jwong101 (changes requested):

crates/forge_loader/src/manifest.rs
gersbach: I hope this fixed it, so an object is the same thing as a mapping in yaml.

Thanks a lot for your help :D

jwong101:
lgtm

PR #55, opened 2024-11-22
Classify functions exposed by Rovo modules as user invocable with testing

dxu2atlassian:
Tests with logic of implementation still being drafted.

### DIFF
--- a/crates/forge_loader/src/manifest.rs
+++ b/crates/forge_loader/src/manifest.rs
@@ -237,6 +237,7 @@ pub struct WorkflowPostFunction<'a> {
     #[serde(flatten, borrow)]
     common_keys: CommonKey<'a>,
 }
+
 // Jira Service Management Modules
 #[derive(Default, Debug, Clone, PartialEq, Eq, Deserialize)]
 #[serde(rename_all = "camelCase")]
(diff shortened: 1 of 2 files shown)


crates/forge_loader/src/manifest.rs
```diff
@@ -249,6 +249,42 @@ struct AssetsImportType<'a> {
...
+    #[serde(rename = "type")]
+    pub input_type: &'a str,
+    pub required: bool,
+    pub description: Option<&'a str>,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize)]
+pub struct Action<'a> {
+    pub key: &'a str,
+    pub name: &'a str, // TODO: this is not in the manifest in the example forge app, wouldn't this fail? https://bitbucket.org/atlassian/rovo-data-discoverability/src/2743f6ca42589c7d1fb2d50e0d9e65e5ef1aebbe/manifest.yml#lines-3
```
jwong101, line 281: You can just leave it off, since `serde` will still deserialize it correctly if an object contains more keys than what you want to deserialize into.

crates/forge_loader/src/manifest.rs
```diff
@@ -249,6 +249,42 @@ struct AssetsImportType<'a> {
...
+    pub function: &'a str,
```
jwong101, line 282: This should probably be an `Option<&'a str>`, since the action can also just have an `endpoint` [property](https://developer.atlassian.com/platform/forge/manifest-reference/modules/rovo-action/#properties) instead of a `function` property.

crates/forge_loader/src/manifest.rs
```diff
@@ -651,7 +709,16 @@ impl<'a> ForgeModules<'a> {
             post.common_keys.append_functions(&mut invokable_functions);
         }
 
-        // JSM modules
+        // Rovo Module Functions
+        // No invokable functions for Rovo Agents but Action can have numerous user invokable functions
+        action.iter().for_each(|action| {
+            if let Some(function) = action.function {
```
jwong101, line 715: You can just directly pass `action.function` to `extend`.

crates/forge_loader/src/manifest.rs
```diff
@@ -249,10 +250,55 @@ struct AssetsImportType<'a> {
...
+    pub description: Option<&'a str>,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize)]
+pub struct Action<'a> {
+    pub key: &'a str,
+
+    // TODO: Rovo Action documentation states this is mandatory field but example apps don't use it, a typo? or are "required" fields not strictly enforced?
+    // eg: https://bitbucket.org/atlassian/rovo-data-discoverability/src/2743f6ca42589c7d1fb2d50e0d9e65e5ef1aebbe/manifest.yml#lines-3
+    pub name: &'a str,
```
jwong101, line 284: If it's not mandatory or you don't know, you can just leave it off, since we really only care about `function`.

jwong101:
LGTM. Clean up the commit history, and I'll merge it in.

PR #56, opened 2024-11-26
EAS-2497 : Collect possible GraphQL queries during lowering

gersbach:


crates/fsrt/src/main.rs
```diff
@@ -99,7 +103,69 @@ impl fmt::Display for Error {
...
+        .filter(|f| permissions_resolver.contains_key(f))
+        .map(|f| permissions_resolver.get(f).unwrap().to_owned())
+        .collect()
+}
+
+fn parse_graphql(s: &str) -> Vec<(&str, &str)> {
+    let mut operations = vec![];
+    if let std::result::Result::Ok(doc) = parse_query::<&str>(s) {
+        doc.definitions.into_iter().for_each(|operation| {
+            if let Definition::Operation(op) = operation {
```
jwong101, line 134: Don't forget to collect possible queries from fragments as well.

crates/fsrt/src/main.rs
```diff
@@ -99,7 +103,69 @@ impl fmt::Display for Error {
...
+                match op {
+                    OperationDefinition::Mutation(mutation) => {
```
jwong101, line 136: You can also have plain selection sets as well that need to be parsed.

crates/fsrt/src/main.rs
```diff
@@ -99,7 +103,69 @@ impl fmt::Display for Error {
...
+    let permissions_resolver: HashMap<(&str, &str), &str> =
+        [(("compass", "searchTeams"), "read:component:compass")]
+            .iter()
+            .cloned()
+            .collect();
+
+    operations
+        .iter()
+        .filter(|f| permissions_resolver.contains_key(f))
+        .map(|f| permissions_resolver.get(f).unwrap().to_owned())
```
jwong101, line 126:
```suggestion
        .filter_map(|f| permissions_resolver.get(f))
```

crates/fsrt/src/main.rs
```diff
@@ -99,7 +103,69 @@ impl fmt::Display for Error {
     }
 }
 
-impl std::error::Error for Error {}
+fn check_graphql_and_perms(val: &Value) -> Vec<&str> {
```
jwong101, line 106: What's the string being returned here? And does it need to be a string in the first place?

crates/fsrt/src/main.rs
gersbach: `fragment componentParts on CompassCatalogQueryApi{
 	searchTeams(input: $test) {
    	... on CompassSearchTeamsConnection{
        nodes {
          teamId
        }
      }
  }
}

query compass_query($test:CompassSearchTeamsInput!) {
  compass {
    ...componentParts
  }
}`

You mean something like this ?

crates/fsrt/src/main.rs
gersbach: Does AGG allow this ?

crates/fsrt/src/main.rs
gersbach: I spent a little while looking at this and I initially tried to do the func mut, but the permissions_resolver.get(f) returns a `&&str` when we need a `&str`. AFAIK, there is no way to convert within the filter_map right?

crates/fsrt/src/main.rs
gersbach: This goes back to my internship, we decided on using string instead of enums for returning permissions.... I suppose we can swtich that. Let me know if we should.

crates/fsrt/src/main.rs
jwong101: Yeah

crates/fsrt/src/main.rs
jwong101: Pretty sure it did last time I checked, there was just a warning in the UI.

crates/fsrt/src/main.rs
jwong101: You can just add a `Option::copied` call after the `get`.

crates/fsrt/src/main.rs
jwong101: The main reason I'm going on this now, is because you have to hardcode the permissions anyway with the mapping, which means that the permissions are known at compile time.

Though if that's a stand in for deserializing an actual permission map where the permissions aren't static, then you can leave it as str, though make sure to add a comment.

gersbach: Yeah, i haven't completely decided on how I'll fetch them. I'll either be parsing the nadel files or just creating a map.... we can discuss this.

crates/fsrt/src/main.rs
gersbach: Added support for this and a test case :)

crates/fsrt/src/main.rs
gersbach: Eventually, I want this to be enums, but I think we should make that change in a different task. Let me know if that sounds good :)

crates/fsrt/src/main.rs
gersbach: Hmmmm... I got an error when I tried, but I added support anyway.... doesn't hurt to.

crates/fsrt/src/main.rs
```diff
@@ -99,7 +105,90 @@ impl fmt::Display for Error {
...
+    match val {
+        Value::Const(Const::Literal(s)) => operations.extend_from_slice(&parse_graphql(s)),
+        Value::Phi(vals) => vals.iter().for_each(|val| match val {
+            Const::Literal(s) => operations.extend_from_slice(&parse_graphql(s)),
+        }),
+        _ => {}
+    }
+    // TODO : Build out permission resolver here
+
+    let permissions_resolver: HashMap<(String, String), &str> = [(
```
jwong101, line 119: I think the first entry `compass` of the tuple key could be an enum of platforms that we support, since this will probably be static even in the final approach.

Also, does the second entry need to be a `String`?

crates/fsrt/src/main.rs
```diff
@@ -99,7 +105,90 @@ impl fmt::Display for Error {
...
+}
+
+fn get_field_and_operation<'a, 'b>(
+    selection: Selection<'a, &'a str>,
+    fragments: &'b HashMap<&'b str, SelectionSet<'a, &'a str>>,
+) -> Vec<(String, String)> {
+    let mut vec = vec![];
+
+    if let Selection::Field(mut type_field) = selection {
+        type_field.selection_set.items.clone().iter().for_each(|f| {
```
jwong101, line 171:
```suggestion
        type_field.selection_set.items.iter().for_each(|f| {
```

crates/fsrt/src/main.rs
gersbach: oh no, this is your day off 😱. sorry

I think it does because when you are iterating through the `doc.definitions` on line 144 it requires you to move the the selection and fragment's ownership into the scope. Then, later you would need to return a reference to data in the fragment potentially, which, at that point, it would be owned by the inner scope. As a result, you need to declare the memory on the heap and pass it back. Unless, I am missing something.

I spent 2 hours trying to do that this morning 💀

crates/fsrt/src/main.rs
gersbach: There is a mutable reference within the iteration (when we replace the FragmentSpread with the Fragment's fields), so I think this needs to be cloned here. Unless I am missing something, which I usually am 🙅

jwong101: Wait, can't you just do it in one pass by iterating over the fields of the fragment spread?

Also, looking at this, you don't have to return `String`s and could just use the `&str`s directly by returning `Vec<(&'a str, &'a str)>`.

crates/fsrt/src/main.rs
jwong101: Ah, can't you just return `Vec<(&str, &str)>` directly from `parse_graphql`? The only reason why you need to convert it to a `String` is because of `get_field_and_operation`.

crates/fsrt/src/main.rs
gersbach: You are right. Only one pass was needed. I removed the second pass. Thanks as always, Josh!

crates/fsrt/src/main.rs
```diff
@@ -99,7 +103,84 @@ impl fmt::Display for Error {
...
+            .into_iter()
+            .collect();
+
+    operations
+        .iter()
+        .filter_map(|f| permissions_resolver.get(f).copied())
+        .collect()
+}
+
+fn parse_graphql(s: &str) -> Vec<(&str, &str)> {
```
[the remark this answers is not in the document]
gersbach: I think I resolved this ????

crates/fsrt/src/main.rs
```diff
@@ -99,7 +103,84 @@ impl fmt::Display for Error {
...
+    let mut operations = vec![];
+    let mut fragments: HashMap<&str, graphql_parser::query::SelectionSet<'_, &str>> =
+        HashMap::new();
+
+    // collect all fragments
+    if let std::result::Result::Ok(doc) = parse_query::<&str>(s) {
+        doc.definitions.clone().into_iter().for_each(|d| {
```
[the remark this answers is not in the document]
gersbach: Fixed that ..... or so I hope

jwong101:
LGTM. Clean up the commit history, and I'll merge it in.

PR #57, opened 2024-12-09
Properly detect Bitbucket product requests with testing

dxu2atlassian:
EAS-2526/EAS-2545: Initial changes imported from other branch post rebase main, compiles and runs but needs test cases.


jwong101:
Looks good so far. Though make sure to remove the `.gitignore` changes.

PR #58, opened 2024-12-16
Jira Service Management permission resolvers with unit tests

dxu2-atlassian:
Draft of JS and JSM permission resolvers. The JSM works as expected and the unit test passes, similar to the Bitbucket addition on the other PR. However, for now, the Jira Software parsing doesn't work as the swagger document doesn't seem to have the "x-atlassian-oauth2-scopes" that is utilized.


gersbach (approved):
lgtm

PR #60, opened 2024-12-18
Jira Software parsing and permission resolvers with unit tests

dxu2atlassian:
As Jira Software swagger doc format isn't the same format as others like Bitbucket, JSM, etc.


crates/forge_permission_resolver/src/permissions_resolver.rs
```diff
@@ -362,4 +390,41 @@ mod test {
...
     }
+
+    #[test]
+    fn test_get_issues_for_epic() {
+        let (permission_map, regex_map) = get_permission_resolver_jira_software();
+        let url = "/rest/agile/1.0/sprint/23";
+        let request_type = RequestType::Get;
+        let result = check_url_for_permissions(&permission_map, &regex_map, request_type, url);
+
+        println!("Permission Map: {:?}", permission_map);
```
gersbach, line 401: @jwong101 @jwong33 should we remove debug text ?

jwong101: Yeah, this should be removed.

gersbach: Yeah, I think that'd be good. I see there is some from my code too 😬 😓

crates/forge_permission_resolver/src/permissions_resolver.rs
```diff
@@ -199,12 +214,25 @@ fn get_request_type(
...
         .permission
         .iter()
         .flat_map(|data| &*data.scopes)
         .cloned()
-        .collect()
+        .collect::<Vec<_>>();
+
+    if scopes.is_empty() {
+        // For Jira Software if the initial scopes are empty, try the scopes from the security field
+        scopes.extend(
```
jwong101, line 226: Are there any cases where the `security` field has additional scopes with a nonempty `permission` field?

crates/forge_analyzer/src/checkers.rs
```diff
@@ -1103,7 +1103,15 @@ impl<'cx> Dataflow<'cx> for PermissionDataflow {
                             first_arg_vec.iter().for_each(|first_arg| {
                                 let first_arg = first_arg.replace(&['\"'][..], "");
                                 second_arg_vec.iter().for_each(|second_arg| {
-                                    if intrinsic_func_type
+                                    if intrinsic_func_type == IntrinsicName::RequestJiraSoftware {
+                                        let permissions = check_url_for_permissions(
+                                            interp.jira_software_permission_resolver,
+                                            interp.jira_software_regex_map,
```
gersbach, line 1109: I think a match statement for `intrinsic_func_type` would be more idiomatic here. I realize this was originally my code, sorry 😓.

Also, maybe we have the second argument be set to get if it does not exists. That eliminates the need for the first if else block.

gersbach: These may be out of scope for this PR, and I can make the changes since this was originally my bad code.

gersbach: @jwong101 @jwong33

PR #61, opened 2024-12-23
EAS-2573 : Add GraphQL support for the permission scanner

gersbach:
This PR will find strings that are not invoked from the regular scan as a conservative safety check.


crates/forge_analyzer/src/definitions.rs
```diff
@@ -3299,6 +3305,23 @@ impl Visit for GlobalCollector<'_> {
     }
 }
 
+struct StringCollector {
+    strings: Vec<Atom>,
+}
+
+impl Visit for StringCollector {
+    fn visit_str(&mut self, n: &Str) {
```
jwong101, line 3313: Should this handle templates as well?

gersbach: Fixed this, and made the vec String instead of Atom

PR #62, opened 2024-12-23
Resolve Jira API requests between JSM/JS/Jira to correct permission map/resolver

dxu2atlassian:
JSM/JS/Jira all seem to be bundled within requestJira inside Forge, this will separate those requests and route them to the proper permission map/resolver for classification.
- https://developer.atlassian.com/cloud/jira/service-desk/rest/intro/#about
- https://developer.atlassian.com/cloud/jira/platform/rest/v2/intro/#about
- https://developer.atlassian.com/cloud/jira/software/rest/intro/#introduction


crates/forge_analyzer/src/definitions.rs
```diff
@@ -1000,15 +1017,35 @@ impl FunctionAnalyzer<'_> {
...
+                                .collect::<String>();
+
+                            resolve_jira_api_type(&url).unwrap_or_else(|| {
+                                warn!("Falling back to generic Jira request");
+                                IntrinsicName::RequestJira // TODO: how should we handle this edge case?
+                            })
+                        }
+                        _ => {
+                            warn!("First parameter to requestJira() is invalid");
+                            IntrinsicName::RequestJira // TODO: how should we handle this edge case?
```
jwong101, line 1040: If we want to produce the least amount of false positives, then we should conservatively assume it can use any of Jira APIs if we can't statically determine which one.

jwong101: Yeah that works.

crates/forge_analyzer/src/definitions.rs
```diff
@@ -991,6 +992,24 @@ impl FunctionAnalyzer<'_> {
             *prop == *"get" || *prop == *"getSecret" || *prop == *"query"
         }
 
+        fn resolve_jira_api_type(url: &str) -> Option<IntrinsicName> {
+            // Pattern matching to classify, eg: api.[asApp | asUser]().requestJira(route`/rest/api/3/myself`);
+            match url {
+                url if url.starts_with("/rest/servicedeskapi/") => {
+                    Some(IntrinsicName::RequestJiraServiceManagement)
+                }
+                url if url.starts_with("/rest/agile/") => Some(IntrinsicName::RequestJiraSoftware),
```
jwong101, line 1001: There are other Jira Software URL patterns like `/rest/devinfo` and `/rest/builds`. See the [API docs](https://developer.atlassian.com/cloud/jira/software/rest/api-group-development-information/#api-rest-devinfo-0-10-bulk-post) for others.

crates/forge_analyzer/src/definitions.rs
```diff
@@ -1006,15 +1038,34 @@ impl FunctionAnalyzer<'_> {
...
+                    match first_arg {
+                        Expr::TaggedTpl(TaggedTpl { tpl, .. }) => {
+                            if let Some(TplElement { raw, .. }) = tpl.quasis.first() {
+                                resolve_jira_api_type(raw).unwrap_or_else(|| {
+                                    // Conservatively assume any of Jira APIs may be used if we can't statically determine which one
+                                    warn!("Falling back to any Jira request");
+                                    IntrinsicName::RequestJiraAny
+                                })
+                            } else {
+                                panic!("No url identifiable to classify requestJira() type");
```
jwong101, line 1055: This should just fallback to `RequestJiraAny` instead of panicing.

crates/forge_analyzer/src/definitions.rs
```diff
@@ -1006,15 +1038,34 @@ impl FunctionAnalyzer<'_> {
...
+                            }
+                        }
+                        _ => {
```
jwong101, line 1058: You should also handle the case for when `first_arg` is a string literal: https://rustdoc.swc.rs/swc_ecma_ast/enum.Expr.html#variant.Lit.

crates/forge_analyzer/src/checkers.rs
```diff
@@ -1112,7 +1119,7 @@ impl<'cx> Dataflow<'cx> for PermissionDataflow {
                     interp.bitbucket_permission_resolver,
                     interp.bitbucket_regex_map,
                 ),
-                _ => unreachable!("Invalid intrinsic function type"),
+                _ => (&PermissionHashMap::new(), &HashMap::<String, Regex>::new()),
```
jwong101, line 1122: Could you add the rest of the cases to make this match exhaustive? So we don't forget to update this in the event that we add other services/intrinsics.

jwong101:
LGTM once the match is exhaustive.

crates/forge_analyzer/src/checkers.rs
jwong101: No, I mean just filling out the match arm with the rest of the variants of `IntrinsicName`. There should be a `rust_analyzer` code action for doing this.

Also, you should install [rust-analyzer](https://rust-analyzer.github.io/) if you haven't already.

PR #63, opened 2025-01-06
EAS-2582 : Map Scopes to OAuth

gersbach:


crates/fsrt/src/main.rs
```diff
@@ -653,10 +687,15 @@ pub(crate) fn scan_directory<'a>(
         used_graphql_perms.extend_from_slice(&graphql_perms_defid);
         used_graphql_perms.extend_from_slice(&graphql_perms_varid);
 
+        let oauth_scopes: Vec<&&String> = used_graphql_perms
+            .iter()
+            .map(|val| scope_name_to_oauth.get(&val.as_str()).unwrap())
```
jwong101, line 692: Would there ever be a case where there's a possible GraphQL perm detected that's not in the scope_name_to_oauth map? If so, we might want to just skip over it, but log a warning instead of crashing.

jwong101: Also, this might be better as a hashset of `&str`, since the Vec might get big if there are a lot of individual api usages.

gersbach: Sounds good. Made this a hashset. I think using `String` is a limitation of the API.

crates/fsrt/src/main.rs
```diff
@@ -653,10 +687,21 @@ pub(crate) fn scan_directory<'a>(
...
 
-        let final_perms: Vec<&String> = perm_interp
+        let oauth_scopes: HashSet<&&String> = used_graphql_perms
+            .iter()
+            .filter_map(|val| {
+                if !scope_name_to_oauth.contains_key(&val) {
+                    warn!("Scope is not contained in the scope definitions")
+                }
+
+                scope_name_to_oauth.get(&val)
```
jwong101, line 697: You can deref the result to get a `HashSet<&str>`.

crates/fsrt/src/main.rs
```diff
@@ -653,10 +687,21 @@ pub(crate) fn scan_directory<'a>(
...
+            })
+            .collect();
+
+        let final_perms: HashSet<&String> = perm_interp
```
jwong101, line 701: Also, we might want to use the `final_perms` to report the vulnerabilities below instead of `perm_interp.permissions`, so that there won't be any duplicate perms.

crates/fsrt/src/main.rs
gersbach: Fixed. Thank you for the help!

crates/fsrt/src/main.rs
gersbach: Fixed. Thank you!

PR #66, opened 2025-02-12
Add support for classifying Compass API calls from @forge/graphql package with tests and testing parameters bug fix 

dxu2-atlassian:
Add Compass asApp/asUser from default import: https://www.npmjs.com/package/@atlassian/forge-graphql

### DIFF
--- a/crates/forge_analyzer/src/definitions.rs
+++ b/crates/forge_analyzer/src/definitions.rs
@@ -621,7 +621,7 @@ enum LowerStage {
     Create,
 }
 
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[derive(Debug, Clone, PartialEq, Eq)]
 pub enum IntrinsicName {
     RequestJiraAny,
     RequestJiraSoftware,
@@ -629,7 +629,7 @@ pub enum IntrinsicName {
     RequestConfluence,
     RequestJira,
     RequestBitbucket,
-    RequestCompass,
+    RequestCompass(String),
     Other,
 }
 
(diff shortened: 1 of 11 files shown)


jwong101:
Looks good so far. Just need to public api methods from that package to scopes

crates/forge_permission_resolver/src/permissions_resolver_compass.rs
```diff
@@ -0,0 +1,70 @@
...
+}
+
+impl Default for CompassPermissionResolver {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+// This is for handling https://developer.atlassian.com/cloud/compass/forge-graphql-toolkit/
+fn load_compass_api_scopes() -> HashMap<String, Vec<String>> {
```
jwong101, line 40:
```suggestion
fn load_compass_api_scopes() -> HashMap<&'static str, Vec<&'static str>> {
```

`str`s created from `include_str!` are `static`, so you don't need to allocate new ones.

crates/forge_analyzer/src/definitions.rs
```diff
@@ -1192,6 +1193,25 @@ impl FunctionAnalyzer<'_> {
                 package.cloned().map(Intrinsic::SecretFunction)
             }
 
+            // graphqlGateway's compass
+            // import graphqlGateway from "@atlassian/forge-graphql";
+            // const { errors, data} = await graphqlGateway.compass.asApp().getComponent({ componentId });
+            // [PropPath::Def(def), PropPath::Static(ref compass), ref auth
+            [PropPath::Def(def), PropPath::Static(ref compass), ref authn @ .., PropPath::Static(ref function_name)]
```
jwong101, line 1200: Couldn't you just match for `PropPath::MemberCall` directly here instead of using a slice pattern?

PR #69, opened 2025-03-09
Add using cached-permissions and cached-permissions-path CLI features

dxu2-atlassian:
Adds the feature to enable the user to use previously cached permission mappings to decrease the runtime. Also has a feature to enable the user to designate where the cache files should be stored, otherwise will use defaults. Documentation is also updated.


crates/forge_permission_resolver/src/permissions_cache.rs
```diff
@@ -0,0 +1,258 @@
...
+
+use crate::permissions_resolver::PermissionHashMap;
+use crate::serde::{SerializablePermissionHashMap, ToStringMap};
+use std::fs::{self, File};
+use std::io::{Read, Write};
+use tracing::debug;
+
+const CACHE_EXPIRATION: Duration = Duration::from_secs(60 * 60 * 24 * 7); // 1 week
+
+fn default_cache_path() -> PathBuf {
```
jwong101, line 15: If `$HOME` isn't set, then we probably shouldn't use the default path.

crates/forge_permission_resolver/src/permissions_cache.rs
```diff
@@ -0,0 +1,258 @@
...
+    }
+
+    fn is_cache_valid(&self, path: &PathBuf) -> bool {
+        if let Ok(metadata) = fs::metadata(path) {
+            if let Ok(modified) = metadata.modified() {
+                if let Ok(duration) = modified.elapsed() {
+                    if duration < self.config.ttl {
+                        return true;
+                    } else {
+                        println!(
```
jwong101, line 68: These should output to `stderr`, since it will clutter the output if someone redirects `fsrt`'s output to a file or program.

jwong101 (changes requested):

crates/forge_permission_resolver/src/permissions_cache.rs
```diff
@@ -0,0 +1,261 @@
...
+        PathBuf::from(home).join(".cache/fsrt")
+    } else {
+        PathBuf::new()
+    }
+}
+
+/// Struct to hold cache-related settings
+#[derive(Default, Debug, Clone, PartialEq, Eq)]
+pub struct CacheConfig {
+    use_cache: bool,
```
jwong101, line 26: You can just treat `None` as not using a cache path. Also, the ttl variable is unnecessary if it's just going to be a constant.

crates/forge_permission_resolver/src/permissions_cache.rs
```diff
@@ -0,0 +1,261 @@
...
+            cache_path: Some(cache_path.unwrap_or_else(default_cache_path)),
+            ttl: CACHE_EXPIRATION,
+        }
+    }
+
+    pub fn use_cache(&self) -> bool {
+        self.use_cache
+    }
+
+    pub fn cache_path(&self) -> &PathBuf {
```
jwong101, line 44:
```suggestion
    pub fn cache_path(&self) -> Option<&Path> {
```

That way, you can get rid of `use_cache()` and avoid bugs by encoding the absence of the cache in the state of the data itself.

crates/forge_permission_resolver/src/permissions_cache.rs
```diff
@@ -0,0 +1,261 @@
...
+
+fn default_cache_path() -> PathBuf {
+    if let Ok(home) = env::var("HOME") {
+        PathBuf::from(home).join(".cache/fsrt")
+    } else {
+        PathBuf::new()
+    }
+}
+
+/// Struct to hold cache-related settings
```
jwong101, line 23: Also, should the `Default` implementation just defer to `default_cache_path`?

jwong101: Also, `default_cache_path()` returns and empty `PathBuf` if there's no home directory. You could change that to return an `Option`, so that we don't try to do lookups when there's no home directory.

crates/forge_permission_resolver/src/permissions_cache.rs
```diff
@@ -0,0 +1,261 @@
...
+            } else {
+                eprintln!("Failed to get modified time for cache file: {:?}", path);
+            }
+        } else {
+            eprintln!("Failed to get metadata for cache file: {:?}", path);
+        }
+        false
+    }
+
+    pub fn read(
```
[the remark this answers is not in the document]
jwong101: Yeah, caching the raw swagger files should be fine, as it's not that big.

crates/forge_permission_resolver/src/permissions_cache.rs
```diff
@@ -0,0 +1,177 @@
...
+                eprintln!("Failed to get modified time for cache file: {:?}", path);
+            }
+        } else {
+            eprintln!("Failed to get metadata for cache file: {:?}", path);
+        }
+        false
+    }
+
+    pub fn read(&self, key: &str) -> Option<String> {
+        if !self.config.use_cache() {
```
jwong101, line 82: Redundant as `get_cache_path` returning `None` implies this.

PR #82, opened 2026-04-06
FSRT=1 by default

gladstone-9:


crates/fsrt/src/main.rs
```diff
@@ -445,7 +445,20 @@ pub(crate) fn scan_directory<'a>(
         .collect::<Vec<_>>();
 
     let config = CacheConfig::new(
-        opts.cached_permissions || std::env::var_os("FSRT_CACHE").is_some_and(|s| !s.is_empty()),
+        opts.cached_permissions
```
jwong101: Conventionally, cli flags that are explicitly set take precedence over environment variables: https://clig.dev/#configuration.

The precedence for using the cache should be:

- flag (if set)
- environment variable (if exists)
- default: true

crates/fsrt/src/main.rs
```diff
@@ -83,7 +83,7 @@ pub struct Args {
     graphql_schema_path: Option<PathBuf>,
 
     /// If true, use cached permissions; otherwise, re-download and reparse Swagger files.
```
jwong101: Maybe change the name to `no_cache` so the default of no option being set can be `false`.

gladstone-9: revised

jwong101 (approved):
LGTM

PR #83, opened 2026-04-10
Basic auth checker

gladstone-9:


crates/forge_analyzer/src/checkers.rs
```diff
@@ -1077,6 +1077,666 @@ impl Checker<'_> for SecretChecker {
...
+pub enum AuthHeaderVulnKind {
+    BasicAuth,
+    BearerAdmin,
+}
+
+#[derive(Debug)]
+pub struct AuthHeaderVuln {
+    kind: AuthHeaderVulnKind,
+    stack: String,
+    entry_func: String,
```
jwong101: Is this the entry function or the actual function where the API call happens?

crates/forge_analyzer/src/checkers.rs
```diff
@@ -1077,6 +1077,666 @@ impl Checker<'_> for SecretChecker {
...
+        self.file
+            .iter()
+            .skip_while(|comp| *comp != "src")
+            .for_each(|comp| comp.hash(&mut hasher));
+        self.entry_func.hash(&mut hasher);
+        self.stack.hash(&mut hasher);
+
+        match self.kind {
+            AuthHeaderVulnKind::BasicAuth => Vulnerability {
+                check_name: format!("Custom-Check-Basic-Authorization-{}", hasher.finish()),
```
jwong101: Let's leave the custom check name hash off, so we can avoid filing another ticket for exemptions.

crates/forge_analyzer/src/checkers.rs
```diff
@@ -1077,6 +1077,666 @@ impl Checker<'_> for SecretChecker {
...
+                description: format!(
+                    "HTTP Basic authentication used in fetch request from {} in {:?}.",
+                    self.entry_func, self.file
+                ),
+                recommendation: "Prefer OAuth or API tokens scoped to least privilege. If Basic auth is required, load credentials from Forge secrets or environment variables and avoid logging or exposing the Authorization header.",
+                proof: format!(
+                    "Basic Authorization header on fetch found via {}",
```
jwong101: It might be better for debugging to have this be the API calls used

crates/forge_analyzer/src/checkers.rs
```diff
@@ -1077,6 +1077,666 @@ impl Checker<'_> for SecretChecker {
...
+/// Cached regex matching Atlassian admin-scoped REST API path patterns.
+fn admin_path_re() -> &'static Regex {
+    static RE: std::sync::OnceLock<Regex> = std::sync::OnceLock::new();
+    RE.get_or_init(|| {
+        Regex::new(
+            r"(?ix)
+              ( ^ | [^A-Za-z0-9_] )
+              /?
+              (
+                  admin/v[12]/orgs                               (?: [^A-Za-z0-9_-] | $ )
```
jwong101, line 1473: Are there any vulns that rely on this? This seems like it might be a common path.

crates/forge_analyzer/src/checkers.rs
gladstone-9: This should follow the same pattern as other scanners where the checker implements the Runner and AuthHeaderVuln implements IntoVuln (specifically for 2 vuln kinds AuthHeaderVulnKind::BasicAuth and AuthHeaderVulnKind::BearerAuth).

crates/forge_analyzer/src/checkers.rs
gladstone-9: Should I remove the check_name field entirely? removed the Custom-Check- prefix for now.

crates/forge_analyzer/src/checkers.rs
gladstone-9: I updated the proof to log the call type and endpoint checked.

crates/forge_analyzer/src/checkers.rs
gladstone-9: This is used to check the endpoint of a Bearer authorization call against patterns derived from the Admin Rest APIs. I believe around 8-10 forge apps use Admin Rest APIs in the findings.

jwong101 (approved):

PR #85, opened 2026-05-14
EAS-4405 - feat: whitelist endpoint for authz scanner

gladstone-9:


jwong101 (approved):
LGTM, though could you add a regression test case to confirm that no AuthZ vulns are filed? Also, squash the formatting commit with the actual changes.

Issue #87, opened 2026-06-03
FSRT silently fails to detect vulnerabilities for valid manifests with given patterns and modules


jvallinas-adaptavist:
Several defects in the source code (`crates/forge_loader/src/manifest.rs` can cause FSRT to **silently return zero vulnerabilities on any Forge app manifest that uses consumer modules, sub-hour scheduled triggers, endpoint-routed triggers, or mapping-style permissions scopes**, all of which are valid and documented Forge manifest features. 

Each bug independently produces the silent no-op, and a real-world app is likely to trigger all four at once.

**Impact**

A known High-severity vulnerability in a Forge app could be completely undetected when the manifest contains any of the four patterns below. In these cases, FSRT exits with `{"count":0,"vulns":[]}` and exit code 0, without any error or warning.

Below the different bug cases and patterns that trigger them:

ID | Pattern that triggers the bug
-- | --
BUG-1 | `consumer` module: direct `function:` key not accepted
BUG-2 | `scheduledTrigger`: `camelCase` interval variants missing from enum (`fiveMinutes`)
BUG-3 | `scheduledTrigger` /  `endpoint:` entries require `function:` field
BUG-4 | `permissions.scopes`: YAML mapping silently rejected (expects sequence)

**Test results**

Each test uses the same minimal manifest + **the same source file with a known vulnerability** that can be flagged by the FSRT tool (in our example, we tried using `@forge/api`’s `api.asApp()` method in a source file). Only one section in the manifest is changed at a time.

Manifest variant | FSRT result
-- | --
Baseline (no problematic sections) | 1 vuln found 🟢 
Baseline + consumer with function: key (BUG-1) | 0 vulns (silently fails validation) 🔴 
Baseline + scheduledTrigger fiveMinute (BUG-2) | 0 vulns (silently fails validation) 🔴 
Baseline + scheduledTrigger endpoint: (BUG-3) | 0 vulns (silently fails validation) 🔴 
Baseline + scopes as YAML mapping (BUG-4) | 0 vulns (silently fails validation) 🔴 

**Steps to reproduce**

Prerequisites: Rust toolchain installed; a Forge source file that contains an authorization vulnerability reachable from a `jira:jqlFunction` handler (for instance, a call to `api.asApp()` without prior authorization to trigger the vulnerability detection).

Install the upstream binary

```
cargo install --git https://github.com/atlassian-labs/FSRT fsrt --locked
```

Create a minimal directory to reproduce the issue

```
[...]

PR #88, opened 2026-06-15
ggladstone/eas 4431

gladstone-9:
### Fix four manifest parsing gaps causing silent zero-finding scans

Tracked in GitHub issue #87 / EAS-4431.

**Changes**
Bug-1 - Consumer function: reference
consumer modules now accept a direct function: field. Previously, Consumer required a resolver, so direct function consumers failed deserialization silently.

Bug-2 - fiveMinute scheduled trigger interval
Adds explicit serde rename for Forge's camelCase fiveMinute interval. The enum's rename_all = "lowercase" would deserialize it as fiveminute, not fiveMinute.

Bug-3 - Scheduled trigger endpoint: routing
scheduledTrigger now accepts endpoint-routed triggers without a function: field. Previously it reused RawTrigger which required function.

Bug-4 - Mapping-style permissions.scopes
permissions.scopes now accepts both sequence and mapping forms. Mapping keys are normalized to scope names, values discarded. Null/bare scopes: continues to deserialize as an empty list.

**Why this matters**
FSRT silently falls back to ForgeManifest::default() on a parse error, scans nothing, and exits cleanly. A real vulnerability is missed with no warning.

**Next Step Actions Needed**
Need to log when FSRT fails so we can update syntax support.

eg. Log manifest parse failures and set errors: true in JSON output so unsupported Forge manifest shapes are observable.

### DIFF
--- a/crates/forge_loader/src/manifest.rs
+++ b/crates/forge_loader/src/manifest.rs
@@ -96,12 +96,15 @@ pub struct Consumer<'a> {
 }
 
 // Trigger Modules
+// https://developer.atlassian.com/platform/forge/manifest-reference/modules/scheduled-trigger/
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Deserialize)]
 #[serde(rename_all = "lowercase")]
 enum Interval {
     Hour,
     Day,
     Week,
+    #[serde(rename = "fiveMinute")]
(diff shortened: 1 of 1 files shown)


gladstone-9:
Immediate bugs resolved. Further investigation of underlying problem in progress.

PR #91, opened 2026-06-18
ggladstone/eas 4431 extend

gladstone-9:
# FSRT Forge Manifest Model & Error Reporting Improvements

## Summary

This PR updates FSRT’s Forge manifest model to handle a set of common, valid manifest shapes that previously caused manifest processing failures and silent scan failures. It also improves reporting so manifest and scan failures are surfaced as structured JSON errors instead of returning empty results or empty stdout.

### 1) Added `error_message` to FSRT reports

- Successful scans now return `error_message: ""`.
- Manifest and scan failures now return `errors: true` with a detailed diagnostic message.

### 2) Stopped silently swallowing manifest parse failures

- Manifest deserialization errors are no longer converted into an empty default manifest.
- FSRT now reports where manifest processing failed, including serde path/location details where available.

### 3) Ensured scan-level errors still produce JSON output

- Errors such as transpiled async scan failures now return a structured error report instead of empty stdout with exit code `0`.

### 4) Expanded manifest parsing support for common valid Forge manifest shapes

FSRT now supports:

- `consumer.function` in addition to the older `consumer.resolver.function` shape.
- `scheduledTrigger.interval: fiveMinute`.
- `scheduledTrigger.endpoint` when no function is specified.
- `trigger.endpoint` when no function is specified.
- Mapping-style `permissions.scopes`.
- Missing top-level `permissions` block, treated as `permissions.scopes: []`.
- Boolean macro config (e.g. `config: true`).
- i18n object titles for `jira:adminPage.title`.
- Owned Rovo agent text fields to support YAML strings that require escape processing, such as newline-containing descriptions/conversation starters.

## Why

Several production manifests used valid Forge manifest patterns that FSRT’s internal manifest model did not support. When those manifests failed to deserialize, FSRT could either silently scan an empty manifest or return no JSON output at all, making failures hard to diagnose in automation.

These changes make FSRT more compatible with real-world Forge manifests and improve diagnostics when FSRT still cannot process a manifest or scan an app.


crates/forge_loader/src/manifest.rs
```diff
@@ -184,18 +187,42 @@ struct ContentByLineItem<'a> {
...
+#[serde(bound(deserialize = "'de: 'a"))]
+enum LocalizedText<'a> {
+    Text(&'a str),
+    I18n { i18n: &'a str },
+}
+
 #[derive(Debug, Clone, PartialEq, Eq, Deserialize, Copy)]
 pub struct JiraAdminPage<'a> {
-    title: &'a str,
+    title: LocalizedText<'a>,
```
jwong101, line 225: We can just delete this, since we only care about deserializing the function.

crates/forge_loader/src/manifest.rs
```diff
@@ -267,15 +294,15 @@ struct AssetsImportType<'a> {
 pub struct RovoAgent<'a> {
     pub key: &'a str,
     pub name: &'a str,
-    pub description: Option<&'a str>,
+    pub description: Option<String>,
```
jwong101, line 297: If these fields are unused, we can just remove them as well.

crates/forge_loader/src/manifest.rs
```diff
@@ -638,10 +665,10 @@ pub struct Module<'a> {
     extra: FxHashMap<String, serde_yaml::Value>,
 }
 
-#[derive(Default, Debug, Clone, Copy, PartialEq, Eq, Deserialize)]
+#[derive(Default, Debug, Clone, PartialEq, Eq, Deserialize)]
 pub struct Resource<'a> {
     pub key: &'a str,
-    pub path: &'a str,
+    pub path: String,
```
jwong101, line 671: Does this need to be an owned `String`?

jwong101 (changes requested):

crates/forge_loader/src/manifest.rs
gladstone-9: There might be another fix, but my understanding is that owned Strings allows serde to deserialize special characters.

This fix was introduced from `failed on resources[4].path: invalid type: string ""static\\issue-activity\\build""` where 

`pub resources: Vec<Resource<'a>>` relies on 

```
pub struct Resource<'a> {
    pub key: &'a str,
    pub path: String,
}
```

so it needs a way to deserialize the special characters ie. `\\`

crates/forge_loader/src/manifest.rs
gladstone-9: Will introduce a more permissive model (either deleting or making fields optional) in another PR

Out of scope of this PR

crates/forge_loader/src/manifest.rs
gladstone-9: Will introduce a more permissive model (either deleting or making fields optional) in another PR

Out of scope of this PR

PR #92, opened 2026-06-18
Ggladstone/release yml

gladstone-9:
Summary
Adds a GitHub Actions release pipeline for FSRT and a release-only Cargo dist profile for smaller published binaries.

Changes
• Added release workflow for tagged v*.. releases
• Supports manual rebuilds via workflow_dispatch with an existing tag_name
• Builds and packages Linux/macOS FSRT binaries
• Publishes/replaces release assets with gh release upload --clobber
• Adds [profile.dist] with stripping, LTO, and single codegen unit

Testing
• Release workflow intended to run on version tags and manual dispatch using an existing tag
• Existing workspace tests run in the release workflow before publishing

Test: Checkout releases [ggladstone/fsrt](https://github.com/gladstone-9/FSRT/releases)

Related: EAS-4428

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -3,7 +3,7 @@ name: release
 on:
   push:
     tags:
-      - "v*.*.*"
+      - "v[0-9]+.[0-9]+.[0-9]+"
   workflow_dispatch:
     inputs:
       tag_name:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,104 @@
+name: release
+
+on:
+  push:
+    tags:
+      - "v*.*.*"
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  test:
+    name: Test workspace
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions-rust-lang/setup-rust-toolchain@v1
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,106 @@
...
+          path: artifacts
+          merge-multiple: true
+
+      - name: Publish release assets
+        env:
+          GH_TOKEN: ${{ github.token }}
+          RELEASE_TAG: ${{ github.event_name == 'workflow_dispatch' && inputs.tag_name || github.ref_name }}
+        run: |
+          gh release view "$RELEASE_TAG" >/dev/null 2>&1 || \
+            gh release create "$RELEASE_TAG" \
```
jwong101, line 101: Let's make this be a draft, so we can sanity check it.

.github/workflows/release.yml
```diff
@@ -0,0 +1,106 @@
+name: release
+
+on:
+  push:
+    tags:
+      - "v*.*.*"
```
jwong101:
```suggestion
      - "v[0-9]+.[0-9]+.[0-9]+"
```

jwong101 (changes requested):

.github/workflows/release.yml
gladstone-9: added `--draft`

PR #97, opened 2026-08-24
mint-fct Environment Key Changes

lhe2-wq:
The branch improves how mint-fct chooses a Forge app installation when GraphQL returns installations from multiple environments.
The resulting selection behavior is:

1. One installation: use it.
2. Explicit environment_key: use the first matching installation.
3. Automatic: use the first production, otherwise first default, otherwise the first returned installation.
4. Explicit key with no match still returns EnvironmentNotFound.


jwong101 (approved):
Lgtm

PR #98, opened 2026-08-24
Mint FIT

lhe2-wq:
**Summary**

- This adds end-to-end support for minting Forge Invocation Tokens (FITs):
  - Exposes remote keys from Forge manifests
  - Adds reusable FIT request and minting APIs to forge_pen_test
  - Adds fsrt mint-fit <MODULE_KEY> <REMOTE_KEY>
  - Supports supplied, cached, or newly minted FCTs
  - Supports custom FCT context with --ctx
  - Adds dry-run output, input validation, tests, and documentation

Note: The cache is not persistent. Look into automatic disk cache or other solutions, but this will probably be another PR.

### DIFF
--- a/crates/forge_pen_test/src/forge_pentester.rs
+++ b/crates/forge_pen_test/src/forge_pentester.rs
@@ -76,11 +76,11 @@ const FIT_MUTATION: &str = r#"mutation SignInvocationTokenForUI($input: SignInvo
   signInvocationTokenForUI(input: $input) {
     forgeInvocationToken {
       jwt
-      expiresAt
     }
   }
 }"#;
 const FIT_OPERATION_NAME: &str = "SignInvocationTokenForUI";
+const REDACTED_FCT: &str = "<FCT selected at runtime>";
 
 #[derive(Debug, Clone, PartialEq, Eq)]
 enum EnvironmentSelection {
@@ -166,11 +166,11 @@ struct FctData {
 struct FctResult {
     success: bool,
     errors: Vec<GraphqlErrorObject>,
-    tokens: Vec<FctToken>,
+    tokens: Vec<ForgeContextToken>,
 }
 
 #[derive(Debug, Deserialize)]
-struct FctToken {
+struct ForgeContextToken {
     jwt: String,
 }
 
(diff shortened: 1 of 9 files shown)


crates/fsrt/src/commands/mint_fit.rs
```diff
@@ -0,0 +1,130 @@
...
+use super::parse_ctx;
+
+const REDACTED_FCT: &str = "<FCT selected at runtime>";
+
+/// `mint-fit` arguments.
+#[derive(Args, Debug)]
+pub(crate) struct MintFitArgs {
+    /// Deployed module key used when an FCT must be minted.
+    #[arg(name = "MODULE_KEY")]
+    module_key: String,
```
jwong101: This should be optional when the FCT is passed in.

crates/fsrt/src/commands/mint_fit.rs
```diff
@@ -0,0 +1,130 @@
...
+    let config = forge_pen_test::FsrtRemoteConfig::from_path(&args.config)?;
+    let tester = forge_pen_test::ForgePenTester::new(&manifest, config)?;
+    if args.dry_run {
+        let request = tester.mint_fit_request(&args.module_key, REDACTED_FCT, &args.remote_key)?;
+        println!("variables={:#}", request.variables);
+        return Ok(());
+    }
+
+    let token = tester.mint_fit(
+        &args.module_key,
```
jwong101: You can simplify this by having this function only accept the remote_key and `fct` and reuse `mint_fct` if no FCT is passed in.

crates/forge_pen_test/src/forge_pentester.rs
```diff
@@ -328,6 +357,101 @@ impl<'manifest, 'source> ForgePenTester<'manifest, 'source> {
...
+        let token = data
+            .result
+            .and_then(|result| result.token)
+            .ok_or(MintError::MissingFitToken)?;
+        debug!("successfully minted Forge Invocation Token");
+        Ok(token.jwt)
+    }
+}
+
+fn validate_remote_key(manifest: &ForgeManifest<'_>, remote_key: &str) -> Result<(), MintError> {
```
jwong101: The command should be permissive and just send the remote key. However, if it fails, print out the possible remotes from the extensions by checking for the `core:remote` extension type.

jwong101 (changes requested):

jwong101:
```rust
  /// Forge remote key.
  #[arg(name = "REMOTE_KEY")]
  remote_key: String,

  /// Module used when minting a new FCT.
  #[arg(
      long = "module",
      value_name = "MODULE_KEY",
      required_unless_present = "fct",
      conflicts_with = "fct"
  )]
  module_key: Option<String>,

  /// Context for a newly minted FCT.
  #[arg(
      long,
      value_parser = parse_ctx,
      requires = "module_key",
      conflicts_with = "fct"
  )]
  ctx: Option<Value>,

  /// Existing FCT.
  #[arg(long, conflicts_with_all = ["module_key", "ctx"])]
  fct: Option<String>,
```
  This produces the intended forms:

```shell
  fsrt mint-fit REMOTE_KEY --fct TOKEN
  fsrt mint-fit REMOTE_KEY --module MODULE_KEY [--ctx JSON]
```

  It rejects:

```shell
  --fct TOKEN --module MODULE_KEY
  --fct TOKEN --ctx '{}'
  --ctx '{}'                       # no module
```

  `ArgGroup::multiple(false)` should not be used across all three because that would also incorrectly
  prohibit --module with --ctx. The runtime API should then accept Option<&str> for the module and require
  it only when it actually needs to mint an FCT.

PR #99, opened 2026-08-24
Basic-block lowering for switch statements, unlabelled break / continue

ddetweiler-blip:
As a prerequisite for implementing generic lowering to SSA from a CFG, rather than the present special-case introduction of PHI nodes for ternaries only. Existing FSRT did not have full support for common forms of break and continue, nor for switch statements. Addressing lowering these common structures to CFGs unblocks generic insertion of phi nodes, replacing the current ad-hoc insertion for ternaries only.

### DIFF
--- a/crates/forge_analyzer/src/checkers.rs
+++ b/crates/forge_analyzer/src/checkers.rs
@@ -851,15 +851,16 @@ impl<'cx> Dataflow<'cx> for SecretDataflow {
 }
 
 pub struct UserAuthZChecker<'a> {
-    suspicious_remotes: &'a [&'a str],
+    _suspicious_remotes: &'a [&'a str],
 }
 
 impl<'a> UserAuthZChecker<'a> {
     pub fn new(ids: &'a [&'a str]) -> Self {
-        Self { suspicious_remotes: ids }
+        Self {
+            _suspicious_remotes: ids,
+        }
     }
 }
-
 
 pub struct SecretChecker {
     vulns: Vec<SecretVuln>,
@@ -2275,5 +2276,3 @@ impl<'cx> Dataflow<'cx> for DefinitionAnalysisRunner {
         }
     }
 }
-
-
--- a/crates/forge_analyzer/src/checkers.rs
+++ b/crates/forge_analyzer/src/checkers.rs
@@ -850,6 +850,17 @@ impl<'cx> Dataflow<'cx> for SecretDataflow {
     }
 }
 
+pub struct UserAuthZChecker<'a> {
+    suspicious_remotes: &'a [&'a str],
+}
+
+impl<'a> UserAuthZChecker<'a> {
+    pub fn new(ids: &'a [&'a str]) -> Self {
+        Self { suspicious_remotes: ids }
+    }
+}
+
+
 pub struct SecretChecker {
     vulns: Vec<SecretVuln>,
 }
@@ -2264,3 +2275,5 @@ impl<'cx> Dataflow<'cx> for DefinitionAnalysisRunner {
         }
     }
 }
+
+
--- a/crates/forge_analyzer/src/checkers.rs
+++ b/crates/forge_analyzer/src/checkers.rs
@@ -850,18 +850,6 @@ impl<'cx> Dataflow<'cx> for SecretDataflow {
     }
 }
 
-pub struct UserAuthZChecker<'a> {
-    _suspicious_remotes: &'a [&'a str],
-}
-
-impl<'a> UserAuthZChecker<'a> {
-    pub fn new(ids: &'a [&'a str]) -> Self {
-        Self {
-            _suspicious_remotes: ids,
-        }
-    }
-}
-
 pub struct SecretChecker {
     vulns: Vec<SecretVuln>,
 }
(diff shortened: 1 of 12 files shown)


jwong101:
```javascript
export function whileBreak(run) {
  let result = 'before-loop';

  while (run) {
    result = 'inside-loop';
    // This should jump to the assignment after the loop, not the loop check.
    break;
  }

  result = 'after-loop';
  return result;
}
```

Main branch:

```ir
IR for whileBreak
Variables:
%0: return value
%1: local definition of result
%2: local definition of run
%3: local definition of result
%4: local definition of result
%5: local definition of result
%6: global ref to result
%7: global ref to run
%8: global ref to result
%9: global ref to result
bb0:
    %6 = "before-loop"
    goto bb1
bb1:
    if (%7) then goto bb3 else goto bb2
bb2:
    %8 = "after-loop"
    _ = "after-loop"
    %0 = %8
    return
bb3:
    %9 = "inside-loop"
    _ = "inside-loop"
    goto bb1
```

PR:

```ir
IR for whileBreak
Variables:
%0: return value
%1: local definition of result
%2: local definition of run
%3: local definition of result
%4: local definition of result
%5: local definition of result
%6: global ref to result
%7: global ref to run
%8: global ref to result
bb0:
    %6 = "before-loop"
    goto bb1
bb1:
    if (%7) then goto bb3 else goto bb2
bb2:
    %8 = "after-loop"
    _ = "after-loop"
    %0 = %8
    return
bb3:
    goto bb1
```

Labelled breaks and continues can be deferred, however, the current PR should preserve the break/continue targets for `while, C-style for, and do {} while`.

crates/forge_analyzer/src/checkers.rs
```diff
@@ -850,6 +850,18 @@ impl<'cx> Dataflow<'cx> for SecretDataflow {
     }
 }
 
+pub struct UserAuthZChecker<'a> {
```
jwong101: Can this be deferred to another PR? You can open a second PR based off this one using stacked PRs: https://docs.github.com/en/pull-requests/how-tos/stacked-pull-requests.

jwong101:
You can use this for testing the PR's IR output vs main's IR output:

```javascript

export function forContinue(run, skip) {
  let updates = 0;

  for (; run; updates = updates + 1) {
    if (skip) {
      // This should jump to the update expression, then to the loop check.
      continue;
    }

    updates = updates + 10;
  }

  return updates;
}

export function doWhileBreak(run, stop) {
  let result = 'before-loop';

  do {
    if (stop) {
      // This should jump to the block after the do...while loop.
      break;
    }

    result = 'inside-loop';
  } while (run);

  result = 'after-loop';
  return result;
}

export function forOfBreak(values) {
  let selected = 'none';

  for (const value of values) {
    selected = value;
    // This should exit only this for...of loop.
    break;
  }

  return selected;
}

```

jwong101:
There's a few more regressions mostly involving nested loops and one CFG crash:

```javascript
// Current control-flow lowering regressions


export function forContinueSkipsUpdate(run, skip) {
  let updates = 0;

  for (; run; updates = updates + 1) {
    if (skip) {
      // This should visit the update expression before checking `run` again.
      // The current IR jumps directly to the loop condition instead.
      continue;
    }

    updates = updates + 10;
  }

  return updates;
}

export function doWhileLeaksBreakTarget(outer, inner) {
  let result = 'before-loop';

  while (outer) {
    do {
      result = 'inside-do-while';
    } while (inner);

    // This should exit the outer while loop. The current IR targets the stale
    // do...while exit block, producing a self-loop.
    break;
  }

  result = 'after-loop';
  return result;
}

export function doWhileLeaksContinueTarget(outer, inner) {
  let result = 'before-loop';

  while (outer) {
    do {
      result = 'inside-do-while';
    } while (inner);

    // This should jump to the outer while condition. The current IR jumps to
    // the stale do...while condition instead.
    continue;
  }

  result = 'after-loop';
  return result;
}

export function forOfSwitchContinueCrashesCfg(values, skip) {
  let result = 'before-loop';

  for (const value of values) {
    switch (skip) {
      case true:
        // This is valid JavaScript and should continue the for...of loop. The
        // current IR emits `goto bb4294967295`, then CFG construction panics.
        continue;
      default:
        result = value;
    }
  }

  return result;
}
```

crates/forge_analyzer/src/definitions.rs
```diff
@@ -2038,9 +2217,34 @@ impl FunctionAnalyzer<'_> {
             Stmt::Labeled(LabeledStmt { label, body, .. }) => {
                 self.lower_stmt(body);
             }
-            // TODO: Lower Break and Continue
-            Stmt::Break(BreakStmt { label, .. }) => {}
-            Stmt::Continue(ContinueStmt { label, .. }) => {}
+            Stmt::Break(BreakStmt { label, .. }) => {
+                if label.is_some() {
+                    eprintln!("Labeled breaks are still unsupported");
```
jwong101: Can these be changed to `warn!`?

ddetweiler-blip:
@atlassian-cla-bot check

jwong101 (approved):
LGTM

PR #103, opened 2026-08-27
Invoke extension additions

lhe2-wq:
**Summary**
- Groups Forge penetration-testing commands under `fsrt dast`.
- Adds --app-id support and centralizes app resolution.
- Separates FCT and FIT minting into dedicated modules.
- Adds --entrypoint to invoke-extension, defaulting to resolver.
- Makes the resolver function key optional for non-resolver entrypoints.
- Sends payloads directly to plain handlers while preserving resolver dispatch.
- Updates CLI documentation and example configuration.
- Edited module_keys given, needed filtering by `core:`

### DIFF
--- a/crates/forge_pen_test/src/app_config.rs
+++ b/crates/forge_pen_test/src/app_config.rs
@@ -66,7 +66,6 @@ impl AppConfig {
             .extensions
             .iter()
             .filter(|extension| extension.extension_type == "core:remote")
-            .filter(|extension| !extension.module_key.is_empty())
             .map(|extension| extension.module_key.clone())
             .collect::<Vec<_>>();
         remote_keys.sort(); // why need to sort and dedup?
@@ -161,17 +160,6 @@ mod tests {
                 .is_err()
         );
     }
-
-    #[test]
-    fn missing_module_key_reports_the_actual_extension_list() {
-        assert!(matches!(
-            config(vec![extension("none", "extension-1")])
-                .extension_for_module_key("missing"),
-            Err(PenTestError::ModuleKeyNotFound { module_key, available })
-                if module_key == "missing" && available == ["none"]
-        ));
-    }
-
     #[test]
     fn duplicate_module_keys_return_the_first_extension() {
         let config = config(vec![
@@ -189,7 +177,7 @@ mod tests {
     }
 
     #[test]
-    fn remote_keys_only_include_unique_nonempty_core_remote_extensions() {
+    fn remote_keys_include_unique_core_remote_extensions() {
         let config = config(vec![
             extension_with_type("remote-b", "extension-1", "core:remote"),
             extension_with_type("panel", "extension-2", "jira:issuePanel"),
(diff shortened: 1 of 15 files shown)


crates/forge_pen_test/src/app_config.rs
```diff
@@ -75,6 +77,25 @@ impl AppConfig {
...
+            })
+            .filter(|extension| !extension.module_key.is_empty())
+            .map(|extension| extension.module_key.clone())
+            .fold(Vec::new(), |mut keys, key| {
+                if !keys.contains(&key) {
+                    keys.push(key);
+                }
+                keys
+            })
+    }
```
jwong101: This should preserve the previous behavior, since there are other invokable `core:*` extension types. Also, if there are duplicate or empty module keys, this should be surfaced, as filtering them would mislead someone investigating a Forge app.

crates/forge_pen_test/src/invoke.rs
```diff
@@ -62,45 +63,74 @@ impl ForgePenTester<'_, '_> {
             return Err(PenTestError::EmptySuppliedFct);
         }
         let extension = self.config().extension_for_module_key(module_key)?;
+        let entry_point = entry_point.trim();
```
jwong101: `entryPoint` isn't required and there's a semantic difference between an empty string and not supplying it. If entryPoint is passed, the CLI should pass it verbatim.

jwong101 (changes requested):

crates/forge_pen_test/src/app_config.rs
jwong101: I should have caught this in the previous review, but this shouldn't filter is_empty() for similar reasons.

crates/forge_pen_test/src/invoke.rs
```diff
@@ -63,8 +63,7 @@ impl ForgePenTester {
             return Err(PenTestError::EmptySuppliedFct);
         }
         let extension = self.config().extension_for_module_key(module_key)?;
-        let entry_point = entry_point.trim();
-        if entry_point.is_empty() {
+        if entry_point == Some("") {
```
jwong101: We should just submit an empty string "" if someone explicitly passes an empty string to entryPoint.

crates/forge_pen_test/src/invoke.rs
```diff
@@ -73,7 +72,7 @@ impl ForgePenTester {
             .map(str::trim)
             .map(|key| key.strip_prefix("resolver.").unwrap_or(key))
             .filter(|key| !key.is_empty());
```
jwong101: Same issue here

jwong101 (changes requested):
Also rebase against main.

crates/forge_pen_test/src/invoke.rs
lhe2-wq: Oh ok! My bad, I misunderstood your comment about the semantic difference. Fixing right now!

PR #105, opened 2026-08-31
Mint Session Cookie for FSRT DAST

lhe2-wq:
## Summary

Adds an opt-in `fsrt dast mint-cookie` command to harvest and save a Forge session cookie through a browser login flow.

- Adds optional `[cookie]` configuration for username, headed mode, and verification wait time.
- Saves the harvested session cookie to the configured `auth.raw_cookie_file`.
- Documents the feature and configuration in the README and example config.
- Keeps cookie-harvesting configuration separate from `ForgePenTester`, which only consumes the saved cookie file.

## Note

Need clarification on whether we want to keep this feature separate.
The command is only for tester accounts.


crates/fsrt/src/commands/mint_cookie.rs
```diff
@@ -0,0 +1,179 @@
...
+        let _ = driver.quit().await;
+
+        let value = result?;
+        std::fs::write(&harvest.output, format!("{COOKIE_NAME}={value}"))?;
+        #[cfg(unix)]
+        {
+            use std::os::unix::fs::PermissionsExt;
+            std::fs::set_permissions(&harvest.output, std::fs::Permissions::from_mode(0o600))?;
+        }
+        println!("Wrote {COOKIE_NAME} to {}", harvest.output.display());
```
jwong101: This should write to stderr.

crates/fsrt/src/commands/mint_cookie.rs
```diff
@@ -0,0 +1,179 @@
...
+        capabilities.add_arg("--headless=new")?;
+    }
+
+    let driver = WebDriver::managed(capabilities).match_local().await?;
+    driver.set_implicit_wait_timeout(config.timeout).await?;
+    Ok(driver)
+}
+
+async fn run_flow(driver: &WebDriver, config: &HarvestConfig) -> Result<String> {
+    driver.goto(LOGIN_URL).await?;
```
jwong101: If you use `id.atlassian.com/login?email={EMAIL}&continue={config.site_url}`, then you don't need to fill in the email and it will automatically go to the site url once it successfully logs in.

Note that you still need to click the continue button for username.

crates/fsrt/src/commands/mint_cookie.rs
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    Err(std::io::Error::other(
+        "could not obtain tenant.session.token; retry with --headed to complete any interactive login step",
+    )
+    .into())
+}
+
+async fn click_continue(driver: &WebDriver, timeout: Duration) -> Result<()> {
+    for selector in ["#login-submit", "button[type='submit']"] {
```
jwong101:
```suggestion
    button[data-testid='login-submit-idf-testid']
```

`data-testid` is more stable than the raw id.

crates/fsrt/src/commands/mint_cookie.rs
```diff
@@ -0,0 +1,179 @@
...
+}
+
+async fn build_driver(config: &HarvestConfig) -> Result<WebDriver> {
+    let mut capabilities = ChromeCapabilities::new();
+    if !config.headed {
+        capabilities.add_arg("--headless=new")?;
+    }
+
+    let driver = WebDriver::managed(capabilities).match_local().await?;
+    driver.set_implicit_wait_timeout(config.timeout).await?;
```
jwong101: This makes each query poll use the implicit wait, so this should probably be capped and set to 0 by default.

crates/fsrt/src/commands/mint_cookie.rs
```diff
@@ -0,0 +1,179 @@
...
+        timeout: Duration::from_secs(DEFAULT_TIMEOUT_SECS),
+        verify_wait: Duration::from_secs(
+            cookie.verify_wait_secs.unwrap_or(DEFAULT_VERIFY_WAIT_SECS),
+        ),
+    };
+
+    println!("Harvesting a session cookie for {}", harvest.site_url);
+    tokio::runtime::Runtime::new()?.block_on(async {
+        let driver = build_driver(&harvest).await?;
+        let result = run_flow(&driver, &harvest).await;
```
jwong101:
```suggestion
        let result = driver.run_and_quit(async move |driver| { // use harvest }).await?;
        // You can drop the `driver.quit().await` call if you use this + this handles panics.
```

jwong101 (changes requested):

crates/fsrt/src/commands/mint_cookie.rs
```diff
@@ -0,0 +1,190 @@
...
+        site_url: config.site.to_string(),
+        output: PathBuf::from(&config.auth.raw_cookie_file),
+        headed: args.headed || cookie.headed.unwrap_or(false),
+        timeout: Duration::from_secs(DEFAULT_TIMEOUT_SECS),
+        verify_wait: Duration::from_secs(
+            cookie.verify_wait_secs.unwrap_or(DEFAULT_VERIFY_WAIT_SECS),
+        ),
+    };
+
+    println!("Harvesting a session cookie for {}", harvest.site_url);
```
jwong101:
```suggestion
    erintln!("Harvesting a session cookie for {}", harvest.site_url);
```

`stdout` is for the actual result of running the program. Logging/messaging the user should be done on `stderr`.

crates/fsrt/src/commands/mint_cookie.rs
```diff
@@ -0,0 +1,190 @@
...
+    driver.goto(login_url.as_str()).await?;
+    click_continue(driver, config.timeout).await?;
+
+    let password = query_clickable(driver, "input[name='password']", config.timeout).await?;
+    type_into(&password, &config.password).await?;
+    click_continue(driver, config.timeout).await?;
+
+    let deadline = std::time::Instant::now() + config.timeout;
+    while std::time::Instant::now() < deadline {
+        if matches!(driver.current_url().await, Ok(url) if !url.as_str().contains("id.atlassian.com"))
```
jwong101: It should also check if the url path doesn't start with `SetCST`, as there are some intermediate redirects to `tenant/SetCST[/initiate]` before the cookie is set.

jwong101 (changes requested):

crates/fsrt/src/commands/mint_cookie.rs
lhe2-wq: I see, shouldn't `println!("Wrote {COOKIE_NAME} to {}", harvest.output.display());` write to `stdout` then? Edit: I guess it can be defined as messaging and there is no meaningful result for running the program.

PR #106, opened 2026-09-02
Extending AuthZChecker to remotes requesting system tokens

ddetweiler-blip:
Remote that request system tokens but not user tokens can be surmised to be behaving as `asApp()` calls internally. This PR extends AuthZChecker to recognize `invokeRemote` calls to these, and perform the same checks to determine if the call was authorized or not. For the purposes of AEC checking, we expect apps to perform authorization logic in a way that is detectable to FSRT.

### DIFF
--- a/crates/forge_analyzer/src/checkers.rs
+++ b/crates/forge_analyzer/src/checkers.rs
@@ -381,11 +381,15 @@ impl<'cx> Runner<'cx> for PrototypePollutionChecker {
 
 pub struct AuthZChecker {
     vulns: Vec<AuthZVuln>,
+    check_remotes: bool,
 }
 
 impl AuthZChecker {
-    pub fn new() -> Self {
-        Self { vulns: vec![] }
+    pub fn new(check_remotes: bool) -> Self {
+        Self {
+            vulns: vec![],
+            check_remotes,
+        }
     }
 
     pub fn into_vulns(self) -> impl IntoIterator<Item = AuthZVuln> {
@@ -396,7 +400,7 @@ impl AuthZChecker {
 
 impl Default for AuthZChecker {
     fn default() -> Self {
-        Self::new()
+        Self::new(false)
     }
 }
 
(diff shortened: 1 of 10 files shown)


crates/forge_analyzer/src/checkers.rs
```diff
@@ -488,20 +518,40 @@ impl<'cx> Runner<'cx> for AuthZChecker {
...
             Intrinsic::Fetch => ControlFlow::Continue(*state),
-            Intrinsic::ApiCall(_) if *state != AuthorizeState::Yes => {
-                let vuln = AuthZVuln::new(interp.callstack(), interp.env(), interp.entry());
-                info!("Found a vuln!");
-                self.vulns.push(vuln);
-                ControlFlow::Break(())
-            }
+            Intrinsic::ApiCall(name) if *state != AuthorizeState::Yes => match name {
+                IntrinsicName::InvokeRemote(remote) => {
+                    if let Some(remote_name) = remote
```
ddetweiler-blip: Technically speaking we only get these `ApiCall(InvokeRemote(_))` ones if the remote name is `Some(_)`, otherwise it would have been lowered as `SafeCall(_)`

crates/forge_analyzer/src/checkers.rs
```diff
@@ -491,20 +516,39 @@ impl<'cx> Runner<'cx> for AuthZChecker {
...
+                            interp.callstack(),
+                            interp.env(),
+                            interp.entry(),
+                            RemoteCall(remote_name.clone()),
+                        );
+                        info!("Found a remote vuln!");
+                        self.vulns.push(vuln);
+                        ControlFlow::Break(())
+                    } else {
+                        panic!("unreachable");
```
jwong101: We should just `warn!` if this happens, since we don't want the program to crash for unknown remotes.

crates/forge_analyzer/src/checkers.rs
```diff
@@ -491,20 +516,39 @@ impl<'cx> Runner<'cx> for AuthZChecker {
...
+                    }
+                }
+                _ => {
```
jwong101: Can this be made exhaustive? Since if there's an additional intrinsic added, there's a high chance that this code might need to be updated.

jwong101 (changes requested):

jwong101 (approved):

PR #108, opened 2026-09-08
feat: opt-in Forge secret storage scanner

mfahad-atlassian:
Reports `storage.setSecret`/`getSecret` (and the `@forge/kvs` equivalents) reachable from a non-admin-invokable entry point with no authorization check in between.

Is opt-in and `Medium` severity, with findings intended for manual triage. Enable with `--scanners secret-storage`, or `CHECK_SECRET_STORAGE=1` where the argument list is fixed. It is the first `Scanner` variant excluded from the default set, via `Args::optin_scanner_enabled`.

## Admin pages

An admin page module's own resolver is **not** reported, since the platform restricts it to admin users. A resolver the admin page **shares** with another module **is** reported and labelled `SharedAdminResolver`, because that restriction no longer applies once another module uses the same resolver.

Sharing is detected at handler granularity, because apps do this with two function keys pointing at one handler rather than one key named by two modules — as in `atlassian-labs/gitlab-for-compass` before #87, where `admin-resolver` and `import-resolver` were both handled by `index.resolver`.

## Loader

`Entrypoint.admin` previously matched only `common_keys.function`, so Custom UI admin pages — which declare their handler under `resolver.function` — were not recognised as admin pages at all. New `Entrypoint.admin_only` separates an admin page's own resolver from a shared one, computed from a snapshot of the invokable set taken before the admin pages contribute to it.

`invokable_functions` itself is unchanged, so the existing scanners see exactly the same entry points as before.


mfahad-atlassian:
@atlassian-cla-bot check

crates/forge_analyzer/src/checkers.rs
```diff
@@ -587,6 +589,263 @@ impl Checker<'_> for AuthZChecker {
     type Vuln = AuthZVuln;
 }
 
+/// How the entry point is reachable, which determines why no authorization check
+/// of the app's own is a problem. Surfaced in the finding so triage can
+/// prioritise the shared-resolver case.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum EntryExposure {
+    /// Registered by a module any user can reach.
+    Invokable,
```
jwong101: If there's no other admin module that exposes this, then they can't really authorize anything.
Just handle the shared admin case for now.

crates/forge_analyzer/src/checkers.rs
```diff
@@ -633,14 +892,21 @@ impl<'cx> Dataflow<'cx> for AuthenticateDataflow {
     ) -> Self::State {
         match *intrinsic {
             Intrinsic::Authorize(_) => initial_state,
-            Intrinsic::Fetch | Intrinsic::EnvRead | Intrinsic::StorageRead => {
+            // Reading a secret out of storage is how webtriggers validate an
+            // incoming request, so it counts as authentication evidence. Writing
+            // one does not.
+            Intrinsic::Fetch
+            | Intrinsic::EnvRead
+            | Intrinsic::StorageRead
```
jwong101: Might be cleaner to just add if it's a secret read as a boolean property, so that all of our Forge Storage related functions is on one intrinsic.

crates/forge_analyzer/src/checkers.rs
```diff
@@ -1793,16 +2064,17 @@ impl<'cx> Runner<'cx> for AuthHeaderChecker {
...
-                    base: Base::Var(varid),
-                    ..
-                })) => match interp.get_value(def, *varid, None) {
-                    Some(Value::Const(Const::Literal(s))) => Some(s.clone()),
-                    Some(Value::Phi(phi)) => phi.iter().map(|Const::Literal(s)| s.clone()).next(),
-                    _ => extract_url_prefix_from_body(interp.body(), *varid),
-                },
-                Some(Operand::Lit(lit)) => convert_lit_to_raw(lit),
-                _ => None,
+                Some(
```
jwong101: Was there a bug with the previous version?

crates/forge_loader/src/manifest.rs
```diff
@@ -777,7 +777,13 @@ pub struct Entrypoint<'a, S = Unresolved> {
     pub function: FunctionRef<'a, S>,
     pub invokable: bool,
     pub web_trigger: bool,
+    /// Registered by an admin page module (`jira:adminPage`, `compass:adminPage`).
```
jwong101: Hmm, could we store the modules/module types where the functions are exposed? That way we can have the properties just be predicate methods that inspect the module types. It would also help make the vuln report clearer.

jwong101: Also, because there are other admin module types like `confluence:globalSettings` + project/space admin modules.

jwong101 (changes requested):

crates/forge_analyzer/src/checkers.rs
mfahad-atlassian: No bug, was just deduping. I reverted the changes!

crates/forge_analyzer/src/checkers.rs
mfahad-atlassian: just to be clear, a resolver that writes a secret with no authz is now out of scope for this scanner?

crates/forge_loader/src/manifest.rs
mfahad-atlassian: Do you know if the platform restriction covers the settings modules too?

crates/forge_analyzer/src/checkers.rs
jwong101: We need better heurestics/analysis to cover  set secret on functions that are only exposed through general modules, since a lot of apps just use set secret for metadata.

However, there are apps that run into the same resolver on both admin + general modules foot gun, so this check would be mainly for that.

crates/forge_loader/src/manifest.rs
jwong101: Yes it should work for settings

jwong101:
Could you rebase your PR? There are some merge conflicts.

PR #112, opened 2026-09-15
[1/5] Preserve call identity, locations, and analysis facts

gladstone-9:
**What**

Extend the lowered IR and `Runner` interface to preserve call-site identity, source locations, and argument bindings. Add reusable assignment indexes and constant-value queries over finalized IR. These are the shared analyzer changes needed by the subsequent import-provenance and taint analyses.

**Why**

The existing lowering discards array values and does not retain the call metadata needed for later API recognition and source attribution. Runner callbacks receive a block ID rather than an instruction location, and call frames record the callee rather than the caller.

Dataflow analyses also need to relate multiple `VarId`s to the same resolved `DefId` without conflating shadowed declarations. Assignment lookup and constant checks should be reusable across analyses and resolver runs, rather than repeatedly scanning bodies.

**How**

- Add `CallFacts` keyed by `Location`, retaining the normalized callee path, root binding, and direct import identity. Record instruction spans and ordered argument definitions with their spans.
- Preserve unresolved global callee bindings and lower arrays into `Rvalue::Array`, including representative results from supported expression-bodied `map` callbacks.
- Add lazy `Body` indexes for assignment locations and binding groups, plus environment-scoped assignment indexing and cached conservative constant proofs. These queries operate on finalized IR, after lowering and variable rewriting.
- Pass caller and instruction location through `Runner` visitors and call frames. Update existing checker signatures and keep array summaries out of the legacy `ValueManager` assignment path.
- Register mock project files in the supplied `SourceMap` so recorded spans resolve against the map used by the analysis.

**Validation**

At `1a518b3`: workspace tests passed (116 passed, one existing ignored test), including assignment-index, binding-identity, and constant-query coverage. `cargo fmt --all --check` and `cargo clippy --workspace --all-targets --locked -- -D warnings` passed. Cargo checks used locked dependencies offline.


crates/forge_analyzer/src/definitions.rs
```diff
@@ -1663,11 +1674,35 @@ impl FunctionAnalyzer<'_> {
             let var = self.body.get_or_insert_global(*def_constructor);
             return Operand::with_var(var);
         }
+
+        // Preserve the identity of unresolved global calls in the IR. These
+        // definitions intentionally remain `Undefined`, so analyses can model
+        // known JavaScript built-ins without confusing them with locally
+        // declared or imported functions of the same name.
+        if let Expr::Ident(ident) = expr
```
jwong101: Creating an unresolved `require` binding here breaks the existing CommonJS import collector. `GlobalCollector` now inserts a definition before `FunctionCollector` tests `sym_to_id(require).is_none()`, so the latter no longer recognizes the secret package:

```js
const jwt = require('jsonwebtoken');
export function run() {
  return jwt.sign({ foo: 'bar' }, 'hardcoded-signing-key');
}
```

This reports hardcoded-secret finding on the base and zero on both this PR and the completed stack. The equivalent ESM import still reports on both. Update the legacy built-in/import recognition alongside the new unresolved-global representation and add a CommonJS regression test.

crates/forge_analyzer/src/definitions.rs
```diff
@@ -1663,11 +1674,35 @@ impl FunctionAnalyzer<'_> {
...
+            && self.res.sym_to_id(ident.to_id(), self.module).is_none()
+        {
+            let def = self.res.get_or_insert_sym(ident.to_id(), self.module);
```
jwong101: Materializing the first native `fetch` changes subsequent callee paths from `PropPath::Unknown` to `PropPath::Def`, but `as_intrinsic` only recognizes the former as native fetch. That silently disables token detection on a later call:

```js
export function run() {
  const health = fetch('https://example.com/health');
  const response = fetch('api.atlassian.com/rest/api/3/issue', {
    headers: { Authorization: 'Basic ' + process.env.API_TOKEN }
  });
  return response;
}
```

The base reports one Critical `ATLASSIAN_API_TOKEN` finding; this PR and the stack report zero. Removing the first call restores detection.

jwong101 (changes requested):

